PR #9, opened 2025-12-18
feat: add configurable streaming read timeout

Kartvya69:
## Summary
Add STREAMING_READ_TIMEOUT config option to prevent premature disconnects during long-running streaming responses (e.g., tool calls, complex reasoning).

## Changes
- Add `STREAMING_READ_TIMEOUT` env var (default: 300s / 5 minutes)
- Use `httpx.Timeout` with separate connect/read timeouts for streaming requests
- Update tests to verify new timeout configuration

## Why
During streaming, the model may pause between chunks while "thinking" (especially for tool calls or complex reasoning). The previous implementation used the same timeout for both connection and read operations, which could cause premature disconnects during these pauses.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,7 @@
+# Contributors
+
+Thank you to all the contributors who have helped improve this project!
+
+## Contributors
+
+- [Kartvya69](https://github.com/Kartvya69)


Kartvya69:

jwadow:

@Kartvya69 Thanks for signing the CLA. I assume DoggyMan67 commits are yours. Please confirm if so - otherwise we may need CLA from all commit authors.

PR #17, opened 2026-01-07
fix(auth): reload SQLite credentials before token refresh

bhaskoro-muthohar:
## Problem

When using AWS SSO OIDC with kiro-cli, the gateway caches credentials at startup. If the user re-logs in via kiro-cli (e.g., after SSO token expiration), the new `refresh_token` is written to SQLite but the gateway keeps using the stale cached token, causing authentication failures.

This requires manually restarting the gateway after every kiro-cli re-login, which is inconvenient.

## Solution

Re-read credentials from SQLite before attempting a token refresh in `_refresh_token_aws_sso_oidc()`. This ensures the gateway picks up fresh tokens after kiro-cli re-login without requiring a restart.

## Changes

- Added 3 lines to `_refresh_token_aws_sso_oidc()` to reload credentials from SQLite before refresh

## Testing

- All 59 existing auth manager tests pass
- Manually verified SQLite reading works correctly

### DIFF
--- a/kiro_gateway/auth.py
+++ b/kiro_gateway/auth.py
@@ -477,21 +477,15 @@ class KiroAuthManager:
         
         logger.info("Refreshing Kiro token via AWS SSO OIDC...")
         
-        # AWS SSO OIDC uses form-urlencoded data
+        # AWS SSO OIDC uses JSON body with camelCase keys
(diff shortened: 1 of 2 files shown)


jwadow:
@cla-bot check

Issue #15, opened 2026-01-06
feature request: provide anthropic compatible API


mzazon:
I know the project is openai gateway, but since kiro has all anthropic models, would be nice to have an anthropic compatible API exposed. I get around this using litellm with a your code in a container and litellm provides this anthropic API.

I may take a stab at it when I have some time but thought to raise the issue if you have thought about it

jwadow:
I carried out a large-scale refactoring, the current version is `2.0-rc.1`, everything works under my conditions, but I don’t want to release it yet. I will wait for user feedback and then I will make a normal release.

mzazon:
> Hi [@jwadow](https://github.com/jwadow) I dont' want to open an new issue just for this so, question: Using clients like OpenCode seems to work, but using Claude Code directly I get "API Error: 422", an the app output shows: " HTTP 400 - POST /v1/messages - Invalid model. Please select a different model to continue. (reason: INVALID_MODEL_ID)". I tried with claude-opus-4-5; claude-sonnet-4-5, no luck. Is it expected?
> 
> EDIT: At a first glance I realized CC actually calls a versioned Haiku model that the kiro gateway doesn't expose, so it fails. I added "claude-haiku-4-5-20251001": "claude-haiku-4.5" to the config file and it started working... for a second. it still crashes with the same 422 as soon as it starts to do real work

FYI not sure if this helps but have had to do a lot of playing to get something working but have not tested native anthropic support in claude code yet I did come up with this for .claude/settings.json, maybe it helps:
{
  "env": {
    "ANTHROPIC_BASE_URL": "http://localhost:8888",
    "ANTHROPIC_MODEL": "claude-sonnet-4-5",
    "ANTHROPIC_SMALL_FAST_MODEL": "claude-haiku-4-5",
    "ANTHROPIC_DEFAULT_OPUS_MODEL": "claude-opus-4-5",
    "ANTHROPIC_DEFAULT_HAIKU_MODEL": "claude-haiku-4-5",
    "ANTHROPIC_DEFAULT_SONNET_MODEL": "claude-sonnet-4-5",
    "CLAUDE_CODE_MAX_OUTPUT_TOKENS": "8000",
    "MAX_THINKING_TOKENS": "2000",
    "CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC": "true",
    "CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS": "1"  },

jwadow:
[...]

PR #18, opened 2026-01-07
fix: add system prompt to encourage tool use

bhaskoro-muthohar:
## Problem

When using the gateway with Claude Code CLI, the Kiro model sometimes responds with brief text suggestions like "check litellm config.yaml" instead of actually calling the available tools to perform the requested actions.

## Root Cause

The Kiro model's behavior differs from Claude's native behavior - it occasionally chooses to describe what it would do rather than actually invoking tools, causing conversations to stop abruptly mid-task.

## Solution

Add a system prompt instruction that explicitly tells the model to use tools rather than describe actions:

```
# Tool Use Instructions

IMPORTANT: When you need to perform actions like reading files, running commands, 
or searching, you MUST use the available tools. Do NOT just describe what you would do - 
actually call the tools to do it. Never respond with suggestions like 'check X' or 
'look at Y' without actually using tools to perform those actions.
```

This prompt is only added when tools are present in the request.

## Testing

Tested with Claude Code CLI - conversations that previously stopped mid-way now correctly invoke tools and continue as expected.

### DIFF
--- a/kiro_gateway/converters.py
+++ b/kiro_gateway/converters.py
@@ -82,6 +82,26 @@ def extract_text_content(content: Any) -> str:
     return str(content)
 
 
+def get_tool_use_system_prompt_addition() -> str:
+    """
+    Generate system prompt addition that encourages tool use.
+    
+    Returns:
+        System prompt addition text
+    """
+    return (
+        "\n\n---\n"
+        "# CRITICAL: Tool Use Required\n\n"
+        "You MUST use tools to perform actions. NEVER respond with text like "
+        "\"let me check\", \"try fetching\", or \"I'll search\" without IMMEDIATELY "
+        "calling the appropriate tool in the same response.\n\n"
+        "WRONG: \"Let me search for that information.\"\n"
+        "RIGHT: [Call the search/fetch/read tool directly]\n\n"
+        "If you need to read a file, call Read. If you need to run a command, call Bash. "
+        "If you need to search, call the search tool. Do NOT describe actions - DO them."
+    )
+
+
 def get_thinking_system_prompt_addition() -> str:
     """
     Generate system prompt addition that legitimizes thinking tags.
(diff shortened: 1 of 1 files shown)


bhaskoro-muthohar:
Closing this PR - further investigation revealed the root cause is a parser streaming bug, not model behavior. The model IS calling tools, but the gateway's streaming parser isn't detecting them during live streaming. Will open a new PR with the correct fix.

jwadow:
> Closing this PR - further investigation revealed the root cause is a parser streaming bug, not model behavior. The model IS calling tools, but the gateway's streaming parser isn't detecting them during live streaming. Will open a new PR with the correct fix.

@bhaskoro-muthohar 
If I got the gist of it right with a quick glance, is the behavior of the CLI model different? Yes, this is a justified thing, because even in the IDE API there is some parameter in the header, I don’t remember which one, like “x-amzn-kiro-agent-mode”: “vibe”, affects the model’s responses. Otherwise, it refuses to respond to anything not related to the code. And I selected the current settings (for the IDE) by trial and error so that it would answer everyday questions. Unfortunately, I don’t have access to the CLI, so I added the latest major features almost blindly, based on the advice of other dudes.

bhaskoro-muthohar:
@jwadow Thanks for the context about `x-amzn-kiro-agent-mode: "vibe"` header - that's useful to know!

Update on my investigation: I traced the issue further and it turns out the gateway streaming parser is fine. The actual bug is in **Claude Code itself** - when using `ANTHROPIC_BASE_URL`, it doesn't parse `stop_reason` from the `message_delta` streaming event.

The gateway correctly returns `finish_reason: "tool_calls"`, and LiteLLM correctly converts it to `stop_reason: "tool_use"` in `message_delta`. But Claude Code ignores it and logs `stop_reason: null`, so tools never execute.

I've filed the bug upstream: https://github.com/anthropics/claude-code/issues/16785

So the native Anthropic format support (issue #15) would still be useful to simplify the proxy chain, but won't fix this issue until Claude Code is patched.

Issue #14, opened 2026-01-06
[Bug]: kiro-cli credentials from within a container


mzazon:
### Gateway Version

v1.0.8

### What happened?

Trying to use a container but map the sqlite database to a volume for the container to use in my litellm stack. 
From docker-compose.yaml
  kiro-gateway:
    build: ./kiro-openai-gateway
    container_name: kiro-gateway
    ports:
      - "8000:8000"
    volumes:
      - "/Users/<USERNAME>/Library/Application Support/kiro-cli/data.sqlite3:/data/data.sqlite3:ro"
    restart: unless-stopped

container starts:

2026-01-06 16:03:40 | INFO     | __main__:<module>:305 - Starting Uvicorn server...
2026-01-06 16:03:40 | INFO     | logging:callHandlers:1706 - Started server process [1]
2026-01-06 16:03:40 | INFO     | logging:callHandlers:1706 - Waiting for application startup.
2026-01-06 16:03:40 | INFO     | main:lifespan:234 - Starting application... Creating state managers.
2026-01-06 16:03:40 | INFO     | kiro_gateway.auth:_load_credentials_from_sqlite:255 - Credentials loaded from SQLite database: /data/data.sqlite3
2026-01-06 16:03:40 | INFO     | kiro_gateway.auth:_detect_auth_type:170 - Detected auth type: AWS SSO OIDC (kiro-cli)
2026-01-06 16:03:40 | INFO     | logging:callHandlers:1706 - Application startup complete.
2026-01-06 16:03:40 | INFO     | logging:callHandlers:1706 - Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
2026-01-06 16:03:43 | INFO     | kiro_gateway.routes:get_models:135 - Request to /v1/models
2026-01-06 16:03:43 | INFO     | kiro_gateway.cache:update:75 - Updating model cache. Found 10 models.
2026-01-06 16:03:43 | INFO     | kiro_gateway.routes:get_models:163 - Received 10 models from API

### Debug Logs

Here is what im getting in debug_logs:

root@0bb84b5361e1:/app/debug_logs# cat app_logs.txt
2026-01-06 16:15:17.671 | DEBUG    | kiro_gateway.converters:inject_thinking_tags:162 | Injecting fake reasoning tags with max_tokens=4000
[...]

PR #22, opened 2026-01-08
fix(auth): reload SQLite credentials before AWS SSO OIDC token refresh

kilhyeonjun:
## Problem

When kiro-cli refreshes tokens and updates the SQLite database, kiro-gateway continues using stale credentials from memory, causing 400 errors from AWS SSO OIDC endpoint.

Users had to manually restart the container after every `kiro login`.

### Error Pattern
```
AWS SSO OIDC refresh failed: status=400
body={"error":"invalid_request","error_description":"Invalid request","reason":null}
```

## Solution

Reload credentials from SQLite at the beginning of `_refresh_token_aws_sso_oidc()` to pick up fresh tokens after kiro-cli re-login.

## Changes

- `kiro_gateway/auth.py`: Add SQLite reload before token refresh (4 lines)
- `tests/unit/test_auth_manager.py`: Add test for reload behavior

## Related

- Addresses #14
- Improvement on closed #17 (keeps correct form-urlencoded format)

## Testing

- [x] New test added and passes
- [x] All existing tests pass (60 passed)

### DIFF
--- a/kiro_gateway/auth.py
+++ b/kiro_gateway/auth.py
@@ -464,6 +464,11 @@ class KiroAuthManager:
             ValueError: If required credentials are not set
             httpx.HTTPError: On HTTP request error
         """
+        # Re-read credentials from SQLite to pick up fresh tokens after kiro-cli re-login
+        if self._sqlite_db:
+            logger.debug("Reloading credentials from SQLite before token refresh...")
+            self._load_credentials_from_sqlite(self._sqlite_db)
+        
         if not self._refresh_token:
             raise ValueError("Refresh token is not set")
         if not self._client_id:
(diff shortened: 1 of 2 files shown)


kil-penguin:
@cla-bot check

jwadow (commented):
Thanks, good fix. The approach of reloading SQLite credentials before token refresh correctly handles the case where kiro-cli updates tokens while the gateway is running in a container.

Code is clean - reuses existing method, minimal changes, appropriate log level. Test covers the main scenario well.

One minor notee: sqlite3 and json imports are inside the test function rather than at the top of the file, which differs from the rest of the project. Not a blocker, just noting for consistency.

PR #68, opened 2026-02-06
feat(auth): add KIRO_CLI_API_REGION env var for kiro-cli API region override

Publicker:
## Summary

Adds support for configuring a separate API region for kiro-cli (AWS SSO OIDC) authentication via the `KIRO_CLI_API_REGION` environment variable.

## Problem

When using kiro-cli credentials (SQLite database), the SSO region (e.g., `us-east-1`) may differ from the API region where the Q service is hosted (e.g., `eu-central-1`). This can result in 403 "User is not authorized" errors.

## Solution

- Added `KIRO_CLI_API_REGION` environment variable to `config.py`
- When set, overrides the API region for kiro-cli auth while preserving the SSO region for token refresh
- By default (empty), uses `KIRO_REGION` for both SSO and API (no behavior change)

## Usage

```bash
# Only needed if experiencing 403 errors with kiro-cli credentials
KIRO_CLI_API_REGION="eu-central-1"
```

## Changes

- `kiro/config.py`: Added `KIRO_CLI_API_REGION` config
- `kiro/auth.py`: Updated `_detect_auth_type()` to apply region override when set
- `.env.example`: Documented the new option

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,6 +57,11 @@ PROXY_API_KEY="my-super-secret-password-123"
 # AWS region (default: us-east-1)
 # KIRO_REGION="us-east-1"
 
+# API region for kiro-cli (AWS SSO OIDC) authentication (optional)
+# Only set this if kiro-cli uses a different API region than the SSO region
+# Leave empty to use KIRO_REGION for both SSO and API (default behavior)
+# KIRO_CLI_API_REGION="eu-central-1"
+
 # ===========================================
 # SERVER SETTINGS
 # ===========================================
--- a/kiro/auth.py
+++ b/kiro/auth.py
@@ -40,6 +40,7 @@ from loguru import logger
 
 from kiro.config import (
     TOKEN_REFRESH_THRESHOLD,
+    KIRO_CLI_API_REGION,
     get_kiro_refresh_url,
     get_kiro_api_host,
     get_kiro_q_host,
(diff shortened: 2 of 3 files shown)


Publicker:
> Thanks for the PR! 🎉
> 
> 
> 
> Please reply once with:
> 
> ```
> ```
> 

veenkumarr:
hey @Publicker, nice catch. i was also facing the same issue and created a similar pr aswell. but since ure addressing it, closed my pr. 👍

Publicker:
@cla-bot check

Publicker:
@jwadow no rush, but I'd appreciate a review when you have time

jwadow (commented):
thanks for identifying this and proposing a solution

i've implemented a more comprehensive approach that solves the same problem:
- auto-detection: the gateway now automatically detects the correct API region from your credentials like the sqlite profile ARN or the region field in the json.
- universal override: the KIRO_API_REGION env var now works for all authentication types, not just kiro-cli.
- graceful fallback: there's an automatic fallback chain so it works out of the box for most users.

your use case with kiro-cli using different SSO and API regions is fully covered by the new implementation. the gateway will auto-detect the API region from your profile ARN, and you can still manually override it with KIRO_API_REGION if you need to.

closing this as superseded by #132 and #133

PR #84, opened 2026-02-27
fix(auth): guarantee SQLite connections are closed via try/finally

abhinav-bhateja:
## Problem

Both `_load_credentials_from_sqlite` and `_save_credentials_to_sqlite` used manual `conn.close()` calls scattered across multiple code paths. If any exception is raised between `sqlite3.connect()` and the final `close()` — for example a `json.JSONDecodeError` on a malformed row, or any unexpected Python error — the connection is silently leaked for the lifetime of the process.

`_save_credentials_to_sqlite` was especially fragile: it had **three separate** `conn.close()` calls in different `if/return` branches, making it easy to miss a path in a future refactor.

## Fix

Wrap the cursor work in `try/finally` blocks so `conn.close()` is unconditionally executed on every exit path, including exceptions and early returns:

```python
conn = sqlite3.connect(str(path))
try:
    cursor = conn.cursor()
    # ... all work ...
finally:
    conn.close()
```

This removes 2 out of 3 redundant `conn.close()` calls in `_save_credentials_to_sqlite` and makes the resource lifetime explicit and leak-proof.

## Test plan

- [ ] Start gateway with `KIRO_CLI_DB_FILE` pointing to a valid SQLite database — credentials load correctly
- [ ] Token refresh persists to the database as before
- [ ] Simulate a malformed SQLite row (e.g. invalid JSON) — connection closes cleanly, error is logged

### DIFF
--- a/kiro/auth.py
+++ b/kiro/auth.py
@@ -223,78 +223,80 @@ class KiroAuthManager:
             if not path.exists():
                 logger.warning(f"SQLite database not found: {db_path}")
                 return
-            
+
             conn = sqlite3.connect(str(path))
-            cursor = conn.cursor()
-            
-            # Try all possible token keys in priority order
-            token_row = None
-            for key in SQLITE_TOKEN_KEYS:
-                cursor.execute("SELECT value FROM auth_kv WHERE key = ?", (key,))
-                token_row = cursor.fetchone()
+            try:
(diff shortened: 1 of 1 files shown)


abhinav-bhateja:

jwadow:
@cla-bot check

PR #87, opened 2026-02-27
fix: use full Kiro payload for accurate prompt token counting

PancakeZik:
## Summary

- **prompt_tokens reported to clients were wildly inaccurate** — Claude Code's `/context` command would show ~15k tokens in the header while the per-category breakdown summed to ~170k. This made it impossible to trust context usage.
- The root cause: prompt tokens were derived from Kiro's `contextUsagePercentage`, which returns unreliable/incorrect values. The formula `(contextUsagePercentage / 100) * max_input_tokens` produced numbers far below actual token counts.
- Fixed by counting tokens from the **complete serialized Kiro request payload** (system prompt + messages + tools + all fields) using tiktoken, instead of relying on Kiro's reported percentage. This matches what actually gets sent to the API.

## Changes

- Replace `request_messages`/`request_tools` parameters with a single pre-counted `prompt_tokens: int` across all streaming functions (OpenAI and Anthropic paths)
- Count tokens from the full `kiro_request_body` JSON in both route handlers, before sending to Kiro
- Remove dependency on `contextUsagePercentage` for token counting
- Update tests to match new function signatures

## Test plan

- [x] All 140 existing tests pass
- [x] Verified with Claude Code `/context` — header total now matches the sum of per-category breakdown
- [x] Tested with both fresh conversations (small context) and large conversations (large context)
- [x] Compared behavior against CLIProxyAPIPlus which uses the same approach (counting full request body)

### DIFF
--- a/kiro/routes_anthropic.py
+++ b/kiro/routes_anthropic.py
@@ -50,7 +50,7 @@ from kiro.streaming_anthropic import (
 )
 from kiro.http_client import KiroHttpClient
 from kiro.utils import generate_conversation_id
-from kiro.tokenizer import count_tools_tokens
+from kiro.tokenizer import count_tokens
 
 # Import debug_logger
 try:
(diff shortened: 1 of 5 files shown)


PancakeZik:

PR #88, opened 2026-03-03
feat: add /v1/messages/count_tokens endpoint for Anthropic API

yzddmr6:
Claude Code calls this endpoint before each request to check conversation size and decide whether to trigger compaction. Without it, the gateway returns 404, Claude Code cannot estimate context usage, and long conversations eventually hit the upstream CONTENT_LENGTH_EXCEEDS_THRESHOLD error (400).

The endpoint builds the full Kiro payload and counts tokens on the serialized JSON using tiktoken, consistent with the token counting approach used in the messages endpoint.

before：
```
 API Error: 400 {"type":"error","error":{"type":"api_error","message":"Model context limit reached. Conversation size exceeds model capacity."}}
```
img

after：
image-20260303142451089

### DIFF
--- a/kiro/routes_anthropic.py
+++ b/kiro/routes_anthropic.py
@@ -50,7 +50,7 @@ from kiro.streaming_anthropic import (
 )
 from kiro.http_client import KiroHttpClient
 from kiro.utils import generate_conversation_id
-from kiro.tokenizer import count_tools_tokens
+from kiro.tokenizer import count_tokens, count_tools_tokens
 
 # Import debug_logger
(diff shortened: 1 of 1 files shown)


yzddmr6:

jwadow:
@cla-bot check

jwadow (commented):
thanks for catching that! we've got the endpoint implemented now with all the architectural fixes, including the estimate_request_tokens using the proper claude correction factor on anthropic-format input. closing this out as it's now in main.

PR #95, opened 2026-03-08
fix(debug): preserve request body for debug middleware

### LINKED ISSUE #94: fix debug middleware consuming request body and hanging chat requests
## Summary
- `DebugLoggerMiddleware` reads `request.body()` before routing `/v1/chat/completions` and `/v1/messages`
- Under `BaseHTTPMiddleware`, this can consume the downstream body stream
- FastAPI then waits for the request body again during validation/parsing, causing chat requests to hang with no response bytes returned

## Impact
- OpenCode/OpenAI-compatible chat requests appear to stall indefinitely
- `/health` and `/v1/models` remain healthy, which makes the failure mode confusing
- Triggered when debug logging is enabled (`DEBUG_MODE=all` or `DEBUG_MODE=errors`)

## Reproduction
1. Start the gateway with debug logging enabled
2. Send `POST /v1/chat/completions` with a JSON body
3. Observe the client timing out with `0 bytes received`
4. Restart with `DEBUG_MODE=off` and observe the same request succeed

## Root Cause
The middleware reads the raw request body for logging but does not correctly preserve downstream ASGI receive semantics. This breaks body parsing in later handlers.

## Proposed Fix
- Replay the cached request body back through `request._receive`
- Preserve EOF semantics for both non-empty and empty request bodies
- Add regression tests covering JSON body parsing and empty body behavior

hookokoko:
## Summary
- replay the cached request body after debug logging so downstream FastAPI handlers can still consume it
- preserve EOF semantics for both non-empty and empty request bodies when debug logging is enabled
- add regression coverage for JSON body parsing and empty body passthrough in `DebugLoggerMiddleware`

## Test Plan
- [x] `python -m pytest tests/unit/test_debug_middleware.py -q`
- [x] Start the gateway with debug logging enabled and verify `POST /v1/chat/completions` streams a response successfully

Closes #94

### DIFF
--- a/kiro/debug_middleware.py
+++ b/kiro/debug_middleware.py
@@ -37,6 +37,7 @@ Flush/discard operations are handled by:
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.requests import Request
 from starlette.responses import Response
+from starlette.types import Message
 from loguru import logger
 
 from kiro.config import DEBUG_MODE
@@ -103,6 +104,28 @@ class DebugLoggerMiddleware(BaseHTTPMiddleware):
             body = await request.body()
             if body:
                 debug_logger.log_request_body(body)
+
+            receive_consumed = False
+
+            async def receive() -> Message:
+                """Replay the cached request body exactly once for downstream consumers."""
+                nonlocal receive_consumed
+
+                if not receive_consumed:
+                    receive_consumed = True
+                    return {
+                        "type": "http.request",
+                        "body": body,
+                        "more_body": False,
+                    }
+
+                return {
+                    "type": "http.request",
+                    "body": b"",
+                    "more_body": False,
+                }
+
+            request._receive = receive
         except Exception as e:
             logger.warning(f"Failed to read request body for debug logging: {e}")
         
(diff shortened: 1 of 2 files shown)


hookokoko:

hookokoko:
The PR head commit has been rewritten and now uses my verified GitHub identity: hookokoko <hookokoko@126.com>. I have also accepted the CLA above. Could you please re-run the CLA check?

jwadow:
@cla-bot check

jwadow (commented):
thanks for the PR. I appreciate the effort, but I'm closing this since I can't reproduce the issue. I've been running the gateway 24/7 with DEBUG_MODE="all" for months handling hundreds of daily requests to both /v1/chat/completions and /v1/messages without any hangs or timeouts. everything works fine, requests complete normally, and streaming is smooth. I also ran your exact reproduction steps with large payloads and different request types, and it all worked perfectly with no 0-byte responses. regarding the implementation, I'm not comfortable patching the private request._receive API for a potentially misdiagnosed issue. Starlette already caches request.body() after the first read, so downstream handlers should be fine. if there really was a body consumption issue, the right way to fix it would be switching to pure ASGI middleware or using FastAPI dependency injection, not monkey-patching private APIs that could break with any Starlette update. I can't merge a fix for a problem I can't reproduce, especially with that approach. if you're still seeing this, it might be something specific to your environment like your Python version, dependencies, or OS. feel free to share more details if you figure out what's different. closing for now.

PR #97, opened 2026-03-13
fix(docker): mount kiro-cli volume as rw and exclude .cache from image

### LINKED ISSUE #78: bug(docker): kiro-cli SQLite volume mounted as read-only breaks token persistence
## Bug Description

When using `KIRO_CLI_DB_FILE` authentication with Docker, the documentation and `docker-compose.yml` recommend mounting the kiro-cli SQLite database as read-only (`:ro`). However, `auth.py` writes refreshed tokens back to the SQLite database via `_save_credentials_to_sqlite()`, which fails silently with:

```
SQLite error saving credentials: attempt to write a readonly database
```

## Impact

- Refreshed tokens are never persisted to disk
- After container restart, stale/expired tokens are loaded → authentication failures
- Downstream services (e.g., Telegram bots) receive `Connection error` or `HTTP 504`

## Affected Files

All documentation recommends `:ro` for kiro-cli SQLite mount:

- `docker-compose.yml` L52
- `README.md` L343
- `AGENTS.md` L200
- `docs/ja/README.md` L343
- `docs/ru/README.md` L343
- `docs/pt/README.md` L343
- `docs/zh/README.md` L343
- `docs/ko/README.md` L343
- `docs/id/README.md` L343
- `docs/es/README.md` L343

## Code Reference

`kiro/auth.py`:
- L446: `_save_credentials_to_sqlite()` — writes refreshed tokens to SQLite
- L629, L756: called after token refresh

## Expected Behavior

kiro-cli SQLite volume should be mounted as `:rw` (or without flag, which defaults to rw) so that refreshed tokens can be persisted.

> Note: `~/.aws/sso/cache` mount should remain `:ro` — the gateway only reads from it.

## Suggested Fix

Change kiro-cli SQLite mount from `:ro` to `:rw` in all documentation and `docker-compose.yml`.

### LINKED ISSUE #83: Docker image is too large because /app/.cache/trivy/trivy.db is copied into runtime image
Hi! Thanks for the great project.

  I noticed the published Docker image is much larger than expected, and it looks like a packaging issue.

  ## What I observed

  - Image: `ghcr.io/jwadow/kiro-gateway:latest`
  - Image size: about **1.17GB**
  - Inside container:
    - `/app` is about **950MB**
    - `/app/.cache/trivy` is about **949MB**
    - `/app/.cache/trivy/db/trivy.db` is about **947MB**

  This seems unrelated to runtime requirements and likely got copied into the image during build (`COPY . .`).

  ## Why this matters

  - Slower pull/deploy times
  - More disk usage on servers
  - Unnecessary transfer cost

  ## Suggested fix

  1. Add cache exclusions to `.dockerignore`, at least:
     - `.cache/`
     - `.cache/trivy/`
  2. Consider whitelist-style copy in Dockerfile instead of `COPY . .`, e.g. only:
     - `main.py`
     - `kiro/`
     - `requirements.txt`
     - required runtime files

Logesh-waran2003:
## Summary

Two related Docker issues fixed in a single PR:

**Fix #78 — kiro-cli SQLite volume should be mounted rw, not ro**

The gateway writes refreshed tokens back to the SQLite database via `_save_credentials_to_sqlite()` (called in `auth.py` after token refresh). Mounting the volume as `:ro` causes silent write failures — refreshed tokens are never persisted, leading to stale/expired tokens after container restart and downstream auth failures.

Changed the kiro-cli mount from `:ro` to rw (default, no flag) in:
- `docker-compose.yml`
- `AGENTS.md`
- `README.md`
- All 7 translated docs (`docs/ja`, `ru`, `pt`, `zh`, `ko`, `id`, `es`)

Note: `~/.aws/sso/cache` mounts are unchanged — those remain `:ro` since the gateway only reads from them.

**Fix #83 — exclude `.cache/` from Docker image**

`COPY . .` in the Dockerfile was copying `.cache/trivy/trivy.db` (~947MB) into the runtime image, inflating it to ~1.17GB. Added `.cache/` to `.dockerignore` to prevent this.

## Test plan
- [ ] Container with `KIRO_CLI_DB_FILE` auth persists refreshed tokens across restarts
- [ ] Docker image size is significantly reduced after adding `.cache/` to `.dockerignore`

Closes #78
Closes #83


Logesh-waran2003:

Logesh-waran2003:
@jwadow — CLA acceptance comment posted above. The bot check is still failing; looks like `Logesh-waran2003` may need to be added to the `.clabot` contributors list manually. Happy to wait — just flagging in case it helps.

jwadow:
@cla-bot check

jwadow:
@cla-bot check

jwadow (commented):
thanks for the PR! checked both issues and the fixes look good. the .dockerignore change is perfect, no reason to ship security scan artifacts in the runtime image. for the sqlite mount, you're right that the :ro flag was breaking token persistence. the gateway defaults to writing refreshed tokens back to the db, so the docs should reflect that. users who want read-only mode can still just add :ro and SQLITE_READONLY=true manually. merging now, thanks for tracking down both issues and linking them properly!

ankitcharolia:
@Logesh-waran2003 could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #98, opened 2026-03-14
Add interactive setup CLI with key generation and region selection

river-walras:
… selection


river-walras:

jwadow:
@cla-bot check

jwadow (commented):
thanks for the contribution

i'm closing this pr because:

1. architectural mismatch: the gateway is meant to stay simple with just a .env file, and an interactive cli setup adds too much complexity.
2. region auto-detection: i've already implemented automatic detection from credentials in #132 and #133, so manual selection is mostly unnecessary now.
3. breaking changes: there's no real justification for requiring python 3.13 or changing how we handle dependencies.
4. out of scope: generating nginx configs belongs in the docs, not inside the codebase.

the project philosophy is all about being a simple, transparent proxy gateway. i appreciate the effort, but this doesn't really align with that direction

PR #105, opened 2026-03-20
fix(anthropic): make max_tokens optional for count_tokens endpoint

juslintek:
## Problem

Claude Code calls `/v1/messages/count_tokens` to track context usage and trigger conversation compaction. This endpoint reuses `AnthropicMessagesRequest` which requires `max_tokens: int`.

Since count_tokens only counts tokens (no generation), Claude Code doesn't send `max_tokens`. Every call fails with 422:

```
Validation error (422): 'max_tokens' Field required
```

This breaks context tracking entirely — Claude Code can't estimate context usage, leading to conversations hitting `CONTENT_LENGTH_EXCEEDS_THRESHOLD` errors.

## Solution

Give `max_tokens` a default value of 4096 in `AnthropicMessagesRequest`, making it optional. The messages endpoint still works identically (clients always send it), while count_tokens no longer requires it.

### Depends on
- #103 (feat: add /v1/messages/count_tokens endpoint)

PR #103, opened 2026-03-20
feat: add /v1/messages/count_tokens endpoint for Anthropic API

juslintek:
## Problem

Claude Code calls `/v1/messages/count_tokens` before each request to check conversation size and decide whether to trigger compaction. Without it, the gateway returns 404, Claude Code cannot estimate context usage, and long conversations eventually hit the upstream `CONTENT_LENGTH_EXCEEDS_THRESHOLD` error (400).

## Solution

Add the endpoint that builds the full Kiro payload and counts tokens on the serialized JSON using tiktoken, consistent with the token counting approach used in the messages endpoint.

### Depends on
- #102 (fix: use full Kiro payload for prompt token counting)

PR #106, opened 2026-03-20
fix(anthropic): support built-in server tools and improve count_tokens response

juslintek:
### 1. Built-in server tools cause 422 errors (25 occurrences in logs)

Anthropic's built-in server tools (`web_search_20250305`, `code_execution_20250522`, `bash_20241022`, `text_editor_20250124`, etc.) don't have `input_schema` — they use a `type` field instead. The gateway's `AnthropicTool` model required `input_schema`, causing:

```
Validation error (422): 'input_schema' Field required
input: {'type': 'web_search_20250305', 'name': 'web_search', 'max_uses': 8}
```

### 2. count_tokens response missing `context_management`

The Anthropic API returns `{"input_tokens": N, "context_management": {"original_input_tokens": N}}` but the gateway only returned `{"input_tokens": N}`. Clients expecting the full response format may break.

## Solution

- Make `AnthropicTool.input_schema` optional and add `type` field with `extra="allow"` to accept any built-in tool fields
- Silently strip server tools (no `input_schema`) in the converter since Kiro API can't handle them — custom tools work as before
- Add `context_management.original_input_tokens` to count_tokens response

### Backward compatible
- Custom tools with `input_schema` work identically
- Messages endpoint unaffected
- Only adds acceptance of previously-rejected requests

### Depends on
- #103 (feat: add /v1/messages/count_tokens endpoint)
- #105 (fix: make max_tokens optional)

### DIFF
--- a/kiro/converters_anthropic.py
+++ b/kiro/converters_anthropic.py
@@ -341,6 +341,10 @@ def convert_anthropic_tools(
     """
     Converts Anthropic tools to unified format.
 
+    Silently skips Anthropic built-in server tools (web_search, code_execution,
+    bash, text_editor, etc.) that have no input_schema, since the Kiro API
+    cannot handle them.
+
     Args:
         tools: List of Anthropic tools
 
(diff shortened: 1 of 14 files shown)


Met4physics:
I guess this is not a good solution for problem 1. It just ignore the problem. Kiro IDE can use the web search tool, so there is a way to use it as api.

juslintek:
Good point @Met4physics. You're right that Kiro IDE supports web search internally.

The issue is that the Kiro **API** uses `toolSpecification` format which requires `name`, `description`, and `inputSchema`. Anthropic's server tools (`web_search_20250305`, `code_execution_20250522`, etc.) don't have `inputSchema` — they're a different kind of tool declaration.

If we pass them as-is, the Kiro API rejects them with a 422. If we fabricate an `inputSchema`, the Kiro API wouldn't know to invoke its internal web search — it would just see an unknown tool.

Stripping is the safe default to prevent 422 errors. But I agree this could be improved:
- If the Kiro API exposes its web search as a tool capability in the future, we could map Anthropic's `web_search_20250305` → Kiro's equivalent
- Or we could add a config flag to pass server tools through for users who want to experiment

Happy to iterate on this if there's a known way to trigger Kiro's web search via the API.

juslintek:

jwadow:
@cla-bot check

jwadow (commented):
i'm closing this for now because of the cla requirements. the bot flagged this over a month ago and some of the contributors still haven't signed it, and i can't review or merge anything until every contributor has signed the agreement. if you want to pick this back up later, just open a new pr once everyone's taken care of the paperwork. thanks

PR #96, opened 2026-03-12
fix(anthropic): support deferred tool_reference blocks in tool_result

StitchC:
## Summary
- add `ToolReferenceContentBlock` support in Anthropic models
- allow `tool_result.content` to accept `tool_reference` and unknown content blocks for forward compatibility
- preserve `tool_reference` values during text extraction as readable markers (`[tool_reference: <name>]`)
- add regression tests for models, converters, and route validation

## Problem
Claude Code can send deferred tool references inside `tool_result.content` (for example `{"type":"tool_reference","tool_name":"Read"}`), which caused request validation to fail with HTTP 422.

## Validation
- Unit tests (targeted): 5 passed
- Docker runtime regression: `/v1/messages` payload with `tool_reference` in `tool_result.content` now returns HTTP 200 (no 422)

PR #90, opened 2026-03-05
fix(models): add tool_reference content block support

bhaskoro-muthohar:
## Summary

- Add `ToolReferenceContentBlock` model for Claude Code v2.1.69+ deferred tools (ToolSearch)
- Include it in `ToolResultContentBlock.content` union and `ContentBlock` union
- Skip `tool_reference` in `extract_text_content()` to avoid polluting text extraction
- Add `extra = "allow"` on `ToolResultContentBlock` for forward compatibility

Fixes #89

## Test plan

- [x] Existing tests pass (327 passed)
- [x] Add unit tests for `ToolReferenceContentBlock` model validation
- [x] Add unit tests for `extract_text_content` skipping `tool_reference` blocks
- [x] Test with Claude Code session using deferred tools (ToolSearch)
image

PR #82, opened 2026-02-22
fix(models): sanitize unknown content block types to prevent 422 errors

XuNeo:
## Summary

- Claude Code's `ToolSearch` (deferred tool loading) returns `tool_reference` content blocks inside `tool_result` messages. These are an internal Claude Code content type not recognized by the Anthropic Messages API schema.
- When the conversation history containing these blocks is forwarded through the gateway, Pydantic validation rejects them with a 422 error, breaking all subsequent API calls in the session.
- This PR adds `model_validator(mode="before")` to `ToolResultContentBlock` and `AnthropicMessage` that converts unrecognized block types (e.g. `tool_reference`) into harmless text blocks before Pydantic validation runs.

## Example

Before (causes 422):
```json
{"type": "tool_reference", "tool_name": "mcp__gdbmcp__gdb_connect"}
```

After (passes validation):
```json
{"type": "text", "text": "[tool_reference: mcp__gdbmcp__gdb_connect]"}
```

## Test plan

- [x] All 147 existing unit tests pass
- [x] Manually verified with Claude Code using `ToolSearch` to load deferred MCP tools — no more 422 errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #108, opened 2026-03-21
feat(models): add tool_reference, server_tool_use, and tool_search_tool_result content blocks

juslintek:
## What

Adds Pydantic models for three content block types used by Anthropic's tool search feature, preventing 422 validation errors.

## Why

Claude Code v2.1.69+ with `ENABLE_TOOL_SEARCH=true` sends `tool_reference` blocks inside `tool_result` content. The Anthropic API also returns `server_tool_use` and `tool_search_tool_result` blocks for server-side tool search invocations. Without these models, Pydantic rejects requests with 422 errors.

## Changes

- Add `ToolReferenceContentBlock` model (`type='tool_reference'`)
- Add `ServerToolUseContentBlock` model (`type='server_tool_use'`)
- Add `ToolSearchResultContentBlock` model (`type='tool_search_tool_result'`)
- Add `ToolReferenceContentBlock` to `ToolResultContentBlock.content` union
- Add all three to `ContentBlock` union
- Add `model_config = {"extra": "allow"}` to `ToolResultContentBlock` (for `cache_control` and future fields)

All models use `extra="allow"` for forward compatibility with new fields.

## Related

Overlaps with #90, #96, #82 which address the same 422 issue with different approaches:
- **#90**: Adds `ToolReferenceContentBlock` only (subset of this PR)
- **#96**: Converts `tool_reference` to text markers in `extract_text_content`
- **#82**: Uses `model_validator` to sanitize all unknown blocks to text

This PR takes a typed-model approach covering all three tool search block types, preserving type information for downstream processing.

## Testing

All 1413 existing tests pass. No behavioral changes — only model acceptance is expanded.

### DIFF
--- a/kiro/models_anthropic.py
+++ b/kiro/models_anthropic.py
@@ -78,20 +78,44 @@ class ToolUseContentBlock(BaseModel):
     input: Dict[str, Any]
 
 
+class ToolReferenceContentBlock(BaseModel):
+    """Tool reference block used by Anthropic's tool search / defer_loading feature."""
+
+    type: Literal["tool_reference"] = "tool_reference"
+    tool_name: str
+    model_config = {"extra": "allow"}
+
+
+class ServerToolUseContentBlock(BaseModel):
+    """Server-side tool use block (e.g., tool_search invocations handled by Anthropic API)."""
+
+    type: Literal["server_tool_use"] = "server_tool_use"
+    model_config = {"extra": "allow"}
+
+
+class ToolSearchResultContentBlock(BaseModel):
+    """Tool search result block returned by Anthropic's tool search feature."""
+
+    type: Literal["tool_search_tool_result"] = "tool_search_tool_result"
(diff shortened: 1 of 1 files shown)


juslintek:

jwadow:
@cla-bot check

jwadow (commented):
thanks for the pr, but i can't accept it for a few reasons. first off, there's a conflict with the existing codebase because `ToolReferenceContentBlock` was already added in pr #90, so this would break things on merge. also, we need tests for all new code per our policy in agents.md, but there's nothing here for these new models. 

beyond that, the implementation is incomplete the models are there but they aren't actually handled in the converters or streaming logic, and there's a naming mismatch since you added `tool_search_tool_result` while the code expects `web_search_tool_result`. actually, the current setup with `model_config = {"extra": "allow"}` should already handle unknown blocks without throwing 422s, so we might not even need these models.

PR #109, opened 2026-03-21
feat(anthropic): support defer_loading tool search expansion

juslintek:
## What

Enables Claude Code's MCP Tool Search feature to work through the gateway by handling `defer_loading` tools and expanding `tool_reference` blocks.

## Why

When Claude Code has `ENABLE_TOOL_SEARCH=true`, it sends MCP tools with `defer_loading: true` instead of including them all in every request. The model then requests specific tools on-demand via `tool_reference` blocks. Without this, the gateway either sends all deferred tools (wasting tokens) or drops them entirely.

## How it works

1. `convert_anthropic_tools()` reads `defer_loading` from each tool, stores it on `UnifiedTool._defer_loading`
2. `anthropic_to_kiro()` separates deferred vs active tools
3. Scans messages for `tool_reference` blocks (both top-level and inside `tool_result` content)
4. Expands referenced deferred tools into the active set
5. Only active tools are sent to Kiro API

Also skips Anthropic built-in server tools (no `input_schema`) that the Kiro API cannot handle.

## Changes

- `converters_core.py`: Add `_defer_loading: bool` field to `UnifiedTool` dataclass
- `converters_anthropic.py`: 
  - `convert_anthropic_tools()`: Read `defer_loading`, skip server tools without `input_schema`
  - `anthropic_to_kiro()`: Defer/expand logic with `tool_reference` scanning

## Dependencies

Depends on #108 (tool search content block models)

## Testing

All 1413 existing tests pass. Verified locally with live Claude Code traffic — logs show:
```
[Tool Search] 9 active, 2 deferred, 0 expanded from tool_reference
```

### DIFF
--- a/kiro/converters_anthropic.py
+++ b/kiro/converters_anthropic.py
@@ -341,11 +341,17 @@ def convert_anthropic_tools(
     """
     Converts Anthropic tools to unified format.
 
+    Silently skips Anthropic built-in server tools (web_search, code_execution,
+    bash, text_editor, tool_search_tool_*, etc.) that have no input_schema,
+    since the Kiro API cannot handle them.
+
+    Preserves ``defer_loading`` flag on each tool for downstream expansion.
+
     Args:
-        tools: List of Anthropic tools
+        tools: List of Anthropic tools from the request.
 
     Returns:
-        List of tools in unified format, or None if no tools
+        List of unified tools, or None if no valid tools found.
     """
     if not tools:
         return None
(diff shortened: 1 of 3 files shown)


juslintek:

jwadow:
@cla-bot check

jwadow (commented):
thanks for the pr. i've done a thorough analysis on this including external research and testing across 7 different scenarios, but i'm going to have to decline it. the main issue is that after testing through the gateway's anthropic endpoint and checking the docs, this adds 87 lines of complex logic that doesn't really provide a practical benefit and just adds technical debt. in my tests, the token savings were only about 2.4 percent, around 75 tokens, and your own logs actually confirm that the expansion logic never triggers since they show 0 expanded from tool_reference. 

the feature doesn't really work in practice because anthropic's tool search is a server-side thing. kiro api doesn't support it, so it never returns the tool_reference blocks that the gateway would need to see in the history to trigger the expansion. since the gateway never sees those blocks from the api, the logic is basically dead code. even if it did work, the cost savings are almost nothing-like a fraction of a cent for every thousand requests - which doesn't justify the complexity.

also, this is a major violation of the testing policy in AGENTS.md. we require comprehensive tests for every logical branch, and this has 87 lines of code with zero tests. there are some technical anti-patterns too, like using __dict__ instead of .model_dump() and some weird naming for public fields.

this just doesn't fit the project's philosophy of being a transparent proxy. we fix api quirks rather than trying to implement client-side versions of server-side features. the current behavior of just ignoring the flag and sending all tools is actually better because it's reliable and transparent, even if there's a tiny token overhead. i'd suggest just adding a docstring in converters_anthropic.py to explain that defer_loading isn't supported due to kiro api limitations. i appreciate the effort and the understanding of the spec, but for future prs we really need those tests and a better cost-benefit ratio before adding this much complexity. thanks again for the contribution though.

PR #110, opened 2026-03-21
feat(tools): compress tool schemas and truncate descriptions

juslintek:
## What

Adds opt-in tool schema compression and description truncation to reduce token usage when many MCP tools are present.

## Why

Claude Code with MCP servers can send 100+ tools per request. Each tool's JSON Schema includes verbose fields (description, title, examples, $comment, markdownDescription) that consume context window tokens but aren't needed for tool call generation. With 134 tools observed in live traffic, this saves thousands of tokens per request.

## How

`compress_json_schema()` recursively strips verbose fields from JSON Schema properties, keeping only structural fields (type, properties, required, enum, items, etc.).

Long tool descriptions are also truncated to a configurable limit.

## Changes

- `config.py`: Add `KIRO_COMPRESS_TOOL_SCHEMAS` (default: true) and `KIRO_TOOL_DESC_MAX_CHARS` (default: 200)
- `converters_core.py`: Add `compress_json_schema()` function, integrate into `convert_tools_to_kiro_format()`

## Config

```bash
# Enable/disable schema compression (default: true)
KIRO_COMPRESS_TOOL_SCHEMAS=true

# Max description length, 0 to disable (default: 200)
KIRO_TOOL_DESC_MAX_CHARS=200
```

## Testing

All 1413 existing tests pass. No dependencies on other PRs — this is independent.

### DIFF
--- a/kiro/config.py
+++ b/kiro/config.py
@@ -298,6 +298,20 @@ DEFAULT_MAX_INPUT_TOKENS: int = 200000
 TOOL_DESCRIPTION_MAX_LENGTH: int = int(os.getenv("TOOL_DESCRIPTION_MAX_LENGTH", "10000"))
 
 # ==================================================================================================
+# Tool Schema Compression
+# ==================================================================================================
+
+# Strip verbose fields (description, title, examples, $comment) from tool input schemas.
+# Reduces token usage significantly when many MCP tools are present.
+# Default: true (enabled)
+KIRO_COMPRESS_TOOL_SCHEMAS: bool = os.getenv("KIRO_COMPRESS_TOOL_SCHEMAS", "true").lower() in ("true", "1", "yes")
+
+# Maximum characters for tool descriptions. Longer descriptions are truncated.
+# Set to 0 to disable truncation.
+# Default: 200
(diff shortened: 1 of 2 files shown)


juslintek:

kosuke-zhang:

Author has been inactive for a long time... Could you fork this repo and maintain it?!

jwadow:
@cla-bot check

jwadow (commented):
thanks for the pr. before we look at this, can you clarify a few things? what exact error or issue are you trying to fix here? like, does the kiro api actually reject requests with long schemas, is this just for performance, or do you have a specific error message/issue to point to? also, have you checked how this affects tool call quality? i'm worried that pulling descriptions from properties might make the model worse at calling tools, so do you have any benchmarks or tests to back this up? also, why make it enabled by default if it might hurt quality - shouldn't it be opt-in? finally, this is missing tests, and we need them for every commit.

i'm closing this for now since it's been open a while and needs more work. if you still think this is needed, please open a new pr with a clear description of the problem you're solving (error messages, issue links, etc), tests for the new functionality, benchmarks showing it doesn't hurt tool call quality, and an explanation of why it should be enabled by default. thanks for understandin

PR #112, opened 2026-03-24
feat(thinking): pass client thinking.budget_tokens through to fake reasoning tags

### LINKED ISSUE #111: Fake reasoning ignores client thinking.budget_tokens — always uses hardcoded FAKE_REASONING_MAX_TOKENS
## Bug Description

The "Fake Reasoning" feature injects `<max_thinking_length>` XML tags into prompts to enable extended thinking for non-native-thinking models. However, the injected value is **always** the hardcoded `FAKE_REASONING_MAX_TOKENS` environment variable (default: 4000), completely ignoring the client's `thinking.budget_tokens` from the OpenAI-compatible request body.

This means clients that send a thinking budget (e.g. `"thinking": {"type": "enabled", "budget_tokens": 10000}`) have **no way to control reasoning depth per-request**.

## Steps to Reproduce

1. Set `FAKE_REASONING_ENABLED=true` and `FAKE_REASONING_MAX_TOKENS=4000` in `.env`
2. Send a request with a custom thinking budget:
   ```bash
   curl -X POST http://localhost:8080/v1/chat/completions \
     -H "Content-Type: application/json" \
     -d '{
       "model": "claude-sonnet-4-20250514",
       "messages": [{"role": "user", "content": "Hello"}],
       "thinking": {"type": "enabled", "budget_tokens": 16000}
     }'
   ```
3. Observe the injected `<max_thinking_length>` tag in debug logs

## Expected Behavior

`<max_thinking_length>16000</max_thinking_length>` (client's requested budget)

## Actual Behavior

`<max_thinking_length>4000</max_thinking_length>` (hardcoded env var value, client budget ignored)

## Root Cause

- `ChatCompletionRequest` in `models_openai.py` does not define a `thinking` field, so the value is silently dropped by Pydantic
- `inject_thinking_tags()` in `converters_core.py` has no parameter to accept a client-provided budget
- `build_kiro_payload()` in both core and openai converters have no way to pass the budget through

## Impact

Any client (IDE, CLI tool, etc.) that relies on per-request thinking budget control gets the same static reasoning depth for every request, regardless of task complexity.

kilhyeonjun:
## Summary

Fixes #111 — The fake reasoning feature now respects the client's `thinking.budget_tokens` from the OpenAI-compatible request body, instead of always using the hardcoded `FAKE_REASONING_MAX_TOKENS` env var.

## Changes

- **`models_openai.py`**: Add optional `thinking: Dict[str, Any]` field to `ChatCompletionRequest`
- **`converters_openai.py`**: Extract `budget_tokens` from `request.thinking` and pass it downstream
- **`converters_core.py`**: Accept optional `max_tokens` param in `inject_thinking_tags()` and `thinking_budget` param in `build_kiro_payload()`, falling back to `FAKE_REASONING_MAX_TOKENS` when not provided

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| Client sends `thinking.budget_tokens: 16000` | `<max_thinking_length>4000</max_thinking_length>` (env default) | `<max_thinking_length>16000</max_thinking_length>` |
| Client sends no thinking field | `<max_thinking_length>4000</max_thinking_length>` | `<max_thinking_length>4000</max_thinking_length>` (unchanged) |

## Testing

Verified locally via Docker with debug logging enabled:
```
DEBUG - Client requested thinking budget: 10000
DEBUG - Injecting fake reasoning tags with max_tokens=10000
```

### DIFF
--- a/kiro/config.py
+++ b/kiro/config.py
@@ -416,6 +416,13 @@ FAKE_REASONING_ENABLED: bool = _FAKE_REASONING_RAW not in ("false", "0", "no", "
 # Default: 4000 tokens
 FAKE_REASONING_MAX_TOKENS: int = int(os.getenv("FAKE_REASONING_MAX_TOKENS", "4000"))
 
+# Maximum budget cap for fake reasoning when client sends thinking.budget_tokens.
+# Fake reasoning uses output tokens (not separate thinking tokens), so large budgets
+# can cause the model to spend all output tokens on reasoning with nothing left for actual content.
+# This caps the client's budget_tokens to prevent that. Set to 0 to disable capping.
+# Default: 10000 tokens
+FAKE_REASONING_BUDGET_CAP: int = int(os.getenv("FAKE_REASONING_BUDGET_CAP", "10000"))
+
 # How to handle the thinking block in responses:
 # - "as_reasoning_content": Extract to reasoning_content field (OpenAI-compatible, recommended)
 # - "remove": Remove thinking block completely, return only final answer
--- a/kiro/converters_core.py
+++ b/kiro/converters_core.py
@@ -40,6 +40,7 @@ from kiro.config import (
     TOOL_DESCRIPTION_MAX_LENGTH,
     FAKE_REASONING_ENABLED,
     FAKE_REASONING_MAX_TOKENS,
+    FAKE_REASONING_BUDGET_CAP,
 )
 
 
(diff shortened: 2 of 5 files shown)


jwadow (commented):
Hey @kilhyeonjun thanks for the PR and catching this bug

Unfortunately cant merge this because theres an architectural issue - youre adding a custom `thinking` field to OpenAI API but OpenAI doesnt actually have that parameter. OpenAI uses `reasoning_effort` with values like "low" "medium" "high" etc for their reasoning models. The `thinking.budget_tokens` thing is Anthropics format not OpenAIs

Since kiro-gateway is a proxy that implements both OpenAI-compatible and Anthropic-compatible APIs we cant just make up new parameters that dont exist in the official specs. If we add custom fields clients wont know about them and it breaks the whole point of being compatible

We went ahead and implemented it properly for both APIs:
- OpenAI side uses `reasoning_effort` (the actual official param) and maps it to token budgets with percentages
- Anthropic side uses `thinking.budget_tokens` (also official)
- Added a configurable cap so people dont accidentally burn all their output tokens on thinking
- Wrote 40+ tests to make sure it all works

The fix is already merged and closes #111. You can check converters_openai.py and converters_anthropic.py to see how we handle both APIs

Really appreciate you spotting this and taking time to submit a PR though, the issue was definitely real and needed fixing

PR #116, opened 2026-03-29
fix(auth): retry on 401 invalid_client by reloading device registration from SQLite

KenEzekiel:
## Problem

When `kiro-cli login` is run while the gateway container is already running, the device registration (client_id/client_secret) in SQLite gets rotated. The gateway's in-memory credentials become stale.

Previously, only HTTP 400 (`invalid_grant` — stale refresh_token) triggered a SQLite reload and retry. But a rotated device registration returns HTTP 401 (`invalid_client` — stale client_secret), which was **not** caught — causing permanent auth failure until the container was restarted.

### Error sequence observed:
1. First refresh attempt → `401 invalid_client` (stale client_secret in memory)
2. Gateway does not retry → raises `ValueError` → returns HTTP 500 to clients
3. All subsequent requests also fail with 500

## Fix

- `_refresh_token_aws_sso_oidc()`: now retries on both `400` and `401` by reloading all credentials (token + device registration) from SQLite
- `get_access_token()` graceful degradation: also handles `401` in the fallback path
- Added test for the 401 retry scenario
- Updated docstrings and existing test descriptions

## Testing

- Added `test_refresh_token_aws_sso_oidc_retries_on_401_invalid_client` — verifies 401 triggers SQLite reload and successful retry
- Existing 500 non-retry test still passes (only 400/401 are retried)

### DIFF
--- a/kiro/auth.py
+++ b/kiro/auth.py
@@ -637,8 +637,9 @@ class KiroAuthManager:
         Used by kiro-cli which authenticates via AWS IAM Identity Center.
         
         Strategy: Try with current in-memory token first. If it fails with 400
-        (invalid_request - token was invalidated by kiro-cli re-login), reload
-        credentials from SQLite and retry once.
+        (invalid_request - token was invalidated by kiro-cli re-login) or 401
+        (invalid_client - device registration rotated by kiro-cli re-login),
+        reload all credentials from SQLite and retry once.
         
         This approach handles both scenarios:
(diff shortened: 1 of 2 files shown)


KenEzekiel:

KenEzekiel:
> Thanks for the PR! 🎉
> 
> 
> 
> Please reply once with:
> 
> ```
> ```
> 

KenEzekiel:

KenEzekiel:

PR #120, opened 2026-04-01
fix: credentials file region must not override Kiro API host

d0zingcat:
When logging into Kiro via an organization (e.g. `us-east-2`), that region is written into the credentials file. `_load_credentials_from_file` was using it to overwrite `_api_host` and `_q_host`, routing all API traffic to `https://q.us-east-2.amazonaws.com` — an endpoint that does not exist.

## Changes

- **`kiro/auth.py` — `_load_credentials_from_file`**: Store the credentials `region` field as `_sso_region` (auth/refresh URL only) instead of overriding `_region`/`_api_host`/`_q_host`. This mirrors the already-correct behaviour in `_load_credentials_from_sqlite`, which has a long-standing comment to the same effect.

```python
# Before — breaks API calls when credentials contain a non-existent region
self._region = data['region']
self._refresh_url = get_kiro_refresh_url(self._region)
self._api_host = get_kiro_api_host(self._region)   # ← sends traffic to q.us-east-2.amazonaws.com
self._q_host   = get_kiro_q_host(self._region)

# After — auth uses credentials region; API host stays on KIRO_REGION (default: us-east-1)
self._sso_region   = data['region']
self._refresh_url  = get_kiro_refresh_url(self._sso_region)
# _api_host / _q_host untouched
```

- **`_load_credentials_from_file` docstring**: Documents that `region` is an SSO/auth region and explicitly states API hosts are governed by `KIRO_REGION` env var.

- **`tests/unit/test_auth_manager.py`**: Three new tests in `TestKiroAuthManagerCredentialsFile` covering the regression scenario (`us-east-2` creds → API stays `us-east-1`), `_sso_region` population, and backward-compat when the `region` field is absent.

### DIFF
--- a/kiro/auth.py
+++ b/kiro/auth.py
@@ -310,7 +310,11 @@ class KiroAuthManager:
         - refreshToken: Refresh token
         - accessToken: Access token (if already available)
         - profileArn: Profile ARN
-        - region: AWS region
+        - region: Auth/SSO region — stored as _sso_region and used only for the auth refresh
+          URL (e.g., https://prod.{region}.auth.desktop.kiro.dev/refreshToken). The Kiro API
+          endpoints (_api_host, _q_host) are controlled exclusively by the KIRO_REGION env var
+          (default: us-east-1) and are NOT updated from this field. This prevents invalid
+          organization login regions (e.g., us-east-2) from breaking API calls.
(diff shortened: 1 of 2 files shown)


d0zingcat:

PR #122, opened 2026-04-01
feat: separate SSO authentication zone and Q API zone configuration

d0zingcat:
add the KIRO_API_REGION environment variable to allow independent configuration of the Q Developer API endpoint region. This resolves DNS failures caused by q.amazonaws.com existing only in specific regions while the SSO region may differ.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -54,8 +54,17 @@ PROXY_API_KEY="my-super-secret-password-123"
 # OPTIONAL
 # ===========================================
 
-# AWS region (default: us-east-1)
+# AWS SSO/auth region (default: us-east-1)
+# This controls the OIDC token refresh endpoint: https://oidc.{region}.amazonaws.com/token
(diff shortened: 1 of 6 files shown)


d0zingcat:

jwadow:
@cla-bot check

jwadow (commented):
thanks for flagging this and for the proposal

i decided to go with a more comprehensive approach instead:
- it auto-detects the API region from credentials, like the sqlite profile ARN or the region field in the json
- it lets you manually override it via KIRO_API_REGION (thanks for the idea!)
- it covers all auth types, not just manual config
- it handles edge cases with a graceful fallback

your pr really helped me think through this. i don't gonna merge this specific version, but i appreciate you bringing the regional config issue to our attention. the fix is coming in the 2.4-dev.6 (#132, #133)

PR #123, opened 2026-04-02
feat: add response_format json_object support

lukasBlckIT:
# Problem

When `response_format: {"type": "json_object"}` is set, kiro-gateway accepts the parameter but does not forward it to the Kiro API. Claude responds with markdown-wrapped JSON (```json ... ```) instead of raw JSON, causing JSON parsing failures in downstream clients (e.g. ValidationError in Honcho).

## Solution

Two-layer fix:

1. **System prompt injection** (`kiro/converters_openai.py`) — when `response_format: json_object` is detected, appends a JSON instruction to the system prompt telling Claude to respond with raw JSON only, no markdown wrappers.

2. **Markdown stripping** (`kiro/routes_openai.py` + new `kiro/json_response_format.py`) — post-processes non-streaming responses to strip any remaining markdown code block wrappers as a fallback.

## Behaviour

- `response_format: json_object` → raw, parseable JSON guaranteed
- `response_format: text` or no `response_format` → completely unaffected
- Streaming: system prompt injection is sufficient per OpenAI spec (chunks are text deltas, not full responses)
- Existing system prompts are preserved and the JSON instruction is appended

## Testing

25/25 test cases passed locally, covering non-streaming, streaming, Swedish unicode characters, tool calling, auth errors, and edge cases.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -14,3 +14,4 @@ Thank you to all the contributors who have helped improve this project!
 - [@PAzter1101](https://github.com/PAzter1101) — Docker containerization with CI/CD (#55)
 - [@Ry-DS](https://github.com/Ry-DS) — Images in tool results support for Anthropic MCP servers (#57)
 - [@saaj](https://github.com/saaj) — Regional endpoint fix for eu-central-1 and other non-us-east-1 regions (#58)
+- [@lukasBlckIT](https://github.com/lukasBlckIT) — response_format json_object support (#123)
--- a/kiro/converters_openai.py
+++ b/kiro/converters_openai.py
@@ -296,6 +296,33 @@ def convert_openai_tools_to_unified(tools: Optional[List[Tool]]) -> Optional[Lis
 # Main Entry Point
 # ==================================================================================================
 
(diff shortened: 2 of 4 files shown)


lukasBlckIT:

lukasBlckIT:

jwadow:
@cla-bot check

jwadow (commented):
hey @lukasBlckIT, thanks for the pr and finding this, but i'm gonna have to decline it after looking at the architecture. main deal is there's no config to toggle this, so it hits every user sending json_object, and we can't risk breaking things for the 1000+ people who don't need this. it's also missing the tests required by our policy, and even though you said they passed locally, they aren't in the pr. plus, it only covers non-streaming responses, so you'd still get inconsistent behavior with streaming. honestly, we need to figure out why this is even an issue for you specifically since it's not a problem for anyone else, is it just honcho acting up? before trying to inject prompts, let's see if the kiro api handles this natively or if the gateway is stripping it. if we really need to emulate this, it's gotta be behind a toggle that's off by default, include a full test suite, and handle both streaming and non-streaming correctly. if you can dig into the root cause and show this is a broader issue, i'm happy to take another look, but as is, this is just too much risk. thanks for understanding.

PR #125, opened 2026-04-07
[codex] fix Anthropic tool_reference tool result compatibility

Snow9666:
## Summary

- accept newer Anthropic `tool_result` nested content blocks instead of rejecting them at request validation time
- normalize `tool_reference`, `document`, and `search_result` style blocks into stable Kiro-compatible text summaries
- add model, converter, and route coverage for `tool_reference` payloads

## Root cause

`kiro-gateway` only allowed `tool_result.content` to contain text or image blocks. Anthropic's current Messages API and official SDKs also allow newer nested blocks such as `tool_reference`, which caused `/v1/messages` requests to fail locally with HTTP 422 before the gateway could process them.

## User impact

Clients that send Anthropic deferred-tool / tool-search style payloads with `tool_reference` blocks no longer get rejected by `kiro-gateway` during request validation.

## Validation

- `.venv/bin/pytest -v tests/unit/test_models_anthropic.py tests/unit/test_converters_anthropic.py tests/unit/test_routes_anthropic.py`
- `.venv/bin/pytest -v`
- rebuilt `kiro-gateway` Docker image and replayed a real `tool_reference` request against `/v1/messages` with HTTP 200

## Notes

Full test suite result in this environment: `1414 passed, 2 failed`.
The two failures are existing config-environment assertions unrelated to this Anthropic compatibility fix:
- `tests/unit/test_config.py::TestServerPortConfig::test_default_server_port_is_8000`
- `tests/unit/test_config.py::TestKiroCliDbFileConfig::test_kiro_cli_db_file_from_environment`

### DIFF
--- a/kiro/converters_anthropic.py
+++ b/kiro/converters_anthropic.py
@@ -24,6 +24,7 @@ This module is an adapter layer that converts Anthropic-specific formats
 to the unified format used by converters_core.py.
 """
 
+import json
 from typing import Any, Dict, List, Optional
 
 from loguru import logger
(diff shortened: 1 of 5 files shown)


jwadow:
Nice work on the comprehensive solution! But we went with #90 for now - it's minimal and solves the immediate problem.

Your sanitization logic for encrypted_content is solid though. If we hit document/search_result blocks later, we'll def look at your approach again.

Thanks! 🚀

PR #127, opened 2026-04-12
feat: add install script and full Anthropic content block type parity

severity1:
## Summary
- Add `install.sh` for one-command setup as a user systemd service (requires `uv`)
- Add full Anthropic content block type parity: `RedactedThinkingContentBlock`, `ServerToolUseContentBlock`, `McpToolUseContentBlock`, `ToolReferenceContentBlock`, `DocumentContentBlock`, `AdvisorToolResultContentBlock`, and `Dict[str, Any]` fallback for unknown types
- Minor `.gitignore` updates (Claude Code lock file, `.venv` path fix)

## Test plan
- [x] Verify `bash install.sh` creates and starts the systemd service
- [x] Confirm streaming requests with new content block types (redacted thinking, MCP tool use, server tool use, documents) pass through without errors
- [x] Validate existing message routing still works for standard text/thinking/tool_use blocks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,9 @@ build/
 debug_logs*/
 debug*.json
 
+# Claude Code
+.claude/
+
 # Project-specific
 _notes/
 requests/
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,7 @@ debug_logs*/
 debug*.json
 
 # Claude Code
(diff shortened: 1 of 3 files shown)


severity1:

severity1:
Closing in favor of a cleaner PR with only the models change.

PR #128, opened 2026-04-12
feat(models): add full Anthropic content block type parity

severity1:
## Summary
- Add missing Anthropic content block types: `RedactedThinkingContentBlock`, `ServerToolUseContentBlock`, `McpToolUseContentBlock`, `ToolReferenceContentBlock`, `DocumentContentBlock`, `AdvisorToolResultContentBlock`
- Add `Dict[str, Any]` fallback in `ContentBlock` and response types for forward compatibility with unknown block types
- Expand `ToolResultContentBlock.content` to accept `ToolReferenceContentBlock` and dict fallback

## Test plan
- [ ] Confirm streaming requests with new content block types (redacted thinking, MCP tool use, server tool use, documents) pass through without errors
- [ ] Validate existing message routing still works for standard text/thinking/tool_use blocks

### DIFF
--- a/kiro/models_anthropic.py
+++ b/kiro/models_anthropic.py
@@ -26,7 +26,6 @@ Anthropic's Messages API specification.
 Reference: https://docs.anthropic.com/en/api/messages
 """
 
-import time
 from typing import Any, Dict, List, Literal, Optional, Union
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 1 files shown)


severity1:

jwadow:
@cla-bot check

jwadow (commented):
thanks for the pr, but i can't accept it as is because of some architectural issues. first, it's missing tests, and we need full coverage according to the paranoid testing philosophy in AGENTS.md. you also added 5 content block types that aren't even used anywhere, which just adds bloat. using Dict[str, Any] in unions also breaks pydantic validation, so we need explicit types there. there's also some confusion between streaming and non-streaming types—things like ServerToolUseContentBlock shouldn't be in the base response model. finally, you're missing docs explaining how these are used. before you resubmit, please double check what the Kiro API actually returns, add proper tests, cut the unused types, separate the streaming/response models

PR #134, opened 2026-04-15
fix(auth): preserve existing SQLite fields during token write-back

### LINKED ISSUE #131: bug(auth): _save_credentials_to_sqlite() drops fields, breaking kiro-cli
## Describe the bug

When kiro-gateway refreshes a token via AWS SSO OIDC and writes it back to the kiro-cli SQLite database (`~/.local/share/kiro-cli/data.sqlite3`), it drops several fields that kiro-cli expects. After the write-back, kiro-cli treats the token as invalid and requires the user to re-login.

## Environment

- kiro-gateway: v2.3
- kiro-cli: latest (Rust binary, AWS SSO OIDC auth)
- Auth method: `KIRO_CLI_DB_FILE` pointing to kiro-cli `data.sqlite3`
- OS: Ubuntu Linux

## Root cause

`_save_credentials_to_sqlite()` in `kiro/auth.py` constructs a minimal token object:

```python
token_data = {
    "access_token": self._access_token,
    "refresh_token": self._refresh_token,
    "expires_at": self._expires_at.isoformat() if self._expires_at else None,
    "region": self._sso_region or self._region,
}
```

But the original kiro-cli token in `auth_kv` contains additional required fields:

```json
{
  "client_id": "...",
  "client_secret": "...",
  "scopes": ["..."],
  "region": "eu-west-1",
  "startUrl": "https://...",
  "accessToken": "...",
  "refreshToken": "...",
  "expiresAt": "...",
  "registrationExpiresAt": "..."
}
```

The write-back overwrites the entire JSON value, losing `client_id`, `client_secret`, `startUrl`, and `registrationExpiresAt`. kiro-cli then fails to refresh its own token because the device registration info is gone.

## Steps to reproduce

1. Start kiro-gateway with `KIRO_CLI_DB_FILE=~/.local/share/kiro-cli/data.sqlite3`
2. Make a request through the gateway (triggers token refresh)
3. Try to use kiro-cli → fails, requires re-login

## Suggested fix

**Option A (recommended):** Load the existing JSON from SQLite first, then merge only the refreshed fields into it before writing back. This preserves all other fields.

```python
cursor.execute("SELECT value FROM auth_kv WHERE key = ?", (self._sqlite_token_key,))
row = cursor.fetchone()
if row:
    existing = json.loads(row[0])
    existing["accessToken"] = self._access_token
    existing["refreshToken"] = self._refresh_token
    existing["expiresAt"] = self._expires_at.isoformat() if self._expires_at else None
    token_json = json.dumps(existing)
```

**Option B:** Add a config option to disable write-back entirely (e.g., `KIRO_SQLITE_READONLY=true`), since the gateway keeps refreshed tokens in memory.

## Current workaround

Patch `_save_credentials_to_sqlite()` to return immediately:

```python
def _save_credentials_to_sqlite(self) -> None:
    logger.debug("Skipping SQLite write-back to preserve kiro-cli token integrity")
    return
```

BoosterY:
## Summary

Fixes #131

`_save_credentials_to_sqlite()` previously constructed a new minimal JSON object with only `access_token`, `refresh_token`, `expires_at`, and `region`, then overwrote the entire `auth_kv` entry. This dropped fields that kiro-cli requires (`client_id`, `client_secret`, `startUrl`, `registrationExpiresAt`), causing kiro-cli to require re-login after the gateway refreshed a token.

## Changes

Replaced the full-overwrite with a **read-modify-write** approach:

1. Load the existing JSON from SQLite
2. Update only the token fields (`accessToken`/`refreshToken`/`expiresAt`)
3. Write back the merged result, preserving all other fields

Also handles both camelCase (`accessToken`, used by kiro-cli) and snake_case (`access_token`) field naming conventions.

## Testing

- Verified that after gateway token refresh, kiro-cli can still authenticate without re-login
- Verified that gateway restart correctly reloads and refreshes tokens from SQLite

### DIFF
--- a/kiro/auth.py
+++ b/kiro/auth.py
@@ -445,77 +445,63 @@ class KiroAuthManager:
     
     def _save_credentials_to_sqlite(self) -> None:
         """
-        Saves updated credentials back to SQLite database.
-        
-        This ensures that tokens refreshed by the gateway are persisted
-        and available after gateway restart or for other processes reading
-        the same SQLite database.
-        
-        Strategy:
-        1. If we know which key we loaded from (_sqlite_token_key), save to that key
-        2. If that fails or key is unknown, try all supported keys as fallback
-        
-        This approach ensures credentials are saved to the correct location
-        regardless of authentication type (social login, AWS SSO OIDC, legacy).
-        
-        Updates the auth_kv table with fresh access_token, refresh_token,
-        and expires_at values after successful token refresh.
+        Saves updated credentials back to SQLite database using a
+        read-modify-write approach to preserve fields that kiro-cli needs
+        (e.g. client_id, client_secret, startUrl, registrationExpiresAt).
+
+        Only the token fields (accessToken/access_token, refreshToken/refresh_token,
+        expiresAt/expires_at) are updated; all other fields in the existing JSON
+        are left intact.
         """
         if not self._sqlite_db:
             return
-        
+
         try:
             path = Path(self._sqlite_db).expanduser()
             if not path.exists():
                 logger.warning(f"SQLite database not found for writing: {self._sqlite_db}")
                 return
-            
-            # Use timeout to avoid blocking if database is locked
+
             conn = sqlite3.connect(str(path), timeout=5.0)
             cursor = conn.cursor()
-            
-            # Prepare token data matching the structure from _load_credentials_from_sqlite
-            token_data = {
-                "access_token": self._access_token,
-                "refresh_token": self._refresh_token,
-                "expires_at": self._expires_at.isoformat() if self._expires_at else None,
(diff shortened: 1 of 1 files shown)


BoosterY:

jwadow:
@cla-bot check

jwadow (commented):
Thanks for looking into this! 

I reviewed the code and while the read-modify-write approach is the right idea, the implementation has some issues that would break existing functionality (missing scopes/region updates, no tests to prevent regressions).

Since I can't reproduce the original issue myself and we haven't had other reports, I'm going to close this for now. If this turns out to be a real problem affecting multiple users, we'll implement a proper fix with full test coverage.

Appreciate you taking the time to investigate this though!

PR #143, opened 2026-04-24
Fix Anthropic logical input token usage

kilhyeonjun:
## Summary

This PR fixes inflated Anthropic-compatible token accounting for Claude Code and similar clients.

It does two related things:

- Adds `/v1/messages/count_tokens` for the Anthropic Messages API surface.
- Keeps Anthropic response `usage.input_tokens` aligned with the logical Anthropic request size instead of replacing it with Kiro `contextUsagePercentage` back-calculation.

The goal is to make the Anthropic compatibility layer report the input size that the client actually sent: `messages + system + caller-provided tools`.

## Problem

Claude Code uses Anthropic-compatible token accounting to decide how much context is currently occupied and when compaction should happen.

Before this change, the gateway could significantly over-report input tokens because Kiro reports context usage as a percentage of the model context window, not as an Anthropic-style input token count.

For example, when Kiro reports a high `contextUsagePercentage`, the gateway derived prompt tokens like this:

```text
total_tokens = contextUsagePercentage × maxInputTokens
prompt_tokens = total_tokens - output_tokens
```

That value is useful as a Kiro-side capacity signal, but it is not the same thing as Anthropic `usage.input_tokens`. For Anthropic-compatible clients, using it as `input_tokens` makes the request appear much larger than the logical API payload.

In a local Claude Code session routed through this gateway, this caused `/context` to report a much larger context footprint than the same logical request through direct Anthropic-compatible accounting.

## Root cause

Kiro exposes context usage as a percentage, while Anthropic clients expect request input token counts.

Those two values have different semantics:

- `contextUsagePercentage`: Kiro-side model context utilization signal.
- `usage.input_tokens`: Anthropic-compatible logical input size for the current request.

The gateway already computes a local logical estimate from the incoming Anthropic request. However, the non-streaming Anthropic response path later overwrote that estimate using `contextUsagePercentage × maxInputTokens`.

This made high Kiro context utilization inflate `usage.input_tokens` even when the incoming Anthropic request was small.

### 1. Add Anthropic `/v1/messages/count_tokens`

The new endpoint estimates tokens directly from the incoming Anthropic request:

- `messages`
- `system`
- caller-provided `tools`

It intentionally does not count gateway-added wrapper metadata such as Kiro payload structure or auto-injected gateway behavior.

### 2. Preserve logical `usage.input_tokens` in Anthropic responses

For non-streaming Anthropic responses, this PR stops replacing the precomputed logical request estimate with the Kiro context percentage-derived value.

The response still uses `contextUsagePercentage` to determine whether the upstream stream completed normally, so truncation/completion detection behavior remains intact.

### 3. Add regression coverage

The tests cover:

- `/v1/messages/count_tokens` is registered and returns `input_tokens`.
- Trivial requests remain small and do not include full Kiro wrapper payload overhead.
- Caller-provided tools are counted.
- Structured system blocks are counted without serialization errors.
- High `contextUsagePercentage` does not inflate Anthropic streaming `message_start.usage.input_tokens`.
- High `contextUsagePercentage` does not inflate Anthropic non-streaming `usage.input_tokens`.

## Compatibility notes

This PR changes only Anthropic-compatible token accounting semantics.

It does not change:

- OpenAI-compatible routes.
- Kiro request conversion.
- Kiro streaming parsing.
- Truncation detection based on missing completion/context usage signals.
- Output token counting.

The practical behavior is that Anthropic-compatible clients receive `usage.input_tokens` matching the logical request estimate instead of a Kiro context window utilization-derived number.

## Local validation

Focused unit tests:

```text
./.venv/bin/python -m pytest tests/unit/test_streaming_anthropic.py tests/unit/test_routes_anthropic.py tests/unit/test_tokenizer.py -q
```

Result:

```text
174 passed, 1 warning
```

The warning is an existing `AsyncMock` resource warning in `tests/unit/test_routes_anthropic.py::TestMessagesOptionalParams::test_accepts_stream_true` and is unrelated to this change.

Runtime validation after Docker rebuild:

```text
docker compose -f docker-compose.yml -f docker-compose.local.yml up -d --build
```

Health check:

```text
GET /health
200 {"status":"healthy", ...}
```

Count tokens sanity check:

```text
POST /v1/messages/count_tokens
200 {"input_tokens":9}
```

## Observed impact

In a local Claude Code session routed through the rebuilt gateway, `/context` dropped from roughly:

```text
137k / 200k tokens (69%)
```

to roughly:

```text
62.7k / 200k tokens (31%)
```

for the same class of gateway-routed startup context. The remaining usage is mostly actual tool/schema/context overhead, while the previous Kiro percentage-derived over-reporting is removed.

🤖 Assisted by [kilhyeonjun-harness](https://github.com/kilhyeonjun/kilhyeonjun-harness)


jwadow (commented):
hey, this is the first pr that actually made both me and the ai lose our minds. i tried using a few different models to get reliable documentation for both endpoints, and the consensus kept coming back that you'd handled anthropic but ignored openai. i took that info and we made some architectural changes, but that somehow broke my openai token math, it switched to a local calculation, which is wrong since we have accurate percentages from kiro. 

honestly, i'm pretty lost on how this is all hanging together. everything contradicts itself, and whenever i try to follow the official docs, especially openai's, our token counting breaks. anthropic is even worse since their api is way more convoluted. even if we managed to sort this out, i couldn't merge this as is because i already pushed my own `count_tokens` implementation, so we've got a massive merge conflict anyway.

i'm not closing the pr yet because i want to see how you approach this architecturally. since i don't use claude code, i'd really value input from a power user like you who can verify that the fix is actually solid.

also we need to make sure the streaming and non-streaming behaviors are consistent too. keep in mind that kiro's `contextUsagePercentage` only gives us post-facto numbers for the whole request, but anthropic loves doing things upfront, which is why we have the `tiktoken*1.15` hack, it’s there for cases where the kiro tokens aren't returned in the request. anyway, i'll be waiting to see how you want to handle this or if you've got any updates.

kilhyeonjun:
Thanks for keeping this open. I rechecked the current state and agree this should not be merged as-is after your own `count_tokens` work landed and the branch drifted into conflict.

My suggested architecture from the Claude Code / Anthropic-compatible client side:

1. Keep two token concepts separate:
   - **logical request input tokens**: what the Anthropic-compatible client asked for and what `/v1/messages/count_tokens` should report up front.
   - **Kiro context usage / actual backend usage**: post-facto server-side usage signal from Kiro, which is better for quota/context percentage but should not overwrite the logical Anthropic request accounting.

2. Response `usage.input_tokens` for Anthropic-compatible responses should prefer the logical Anthropic request size when the client expects Anthropic semantics. Kiro's post-facto `contextUsagePercentage` can still be retained separately for internal telemetry / context pressure, but mixing it into Anthropic `usage.input_tokens` makes Claude Code-style clients see confusing token math.

3. Streaming and non-streaming should share the same usage decision path. The streaming path should emit the same final usage semantics as non-streaming, rather than falling back to a separate local estimate.

4. The `tiktoken * 1.15` fallback still makes sense only as a fallback for cases where neither a reliable Anthropic count nor Kiro usage is available. It should not be the primary path once count_tokens / logical sizing is available.

Given the current conflict, I think the cleanest path is for me to rebase/rewrite this PR against your latest implementation rather than trying to preserve this exact diff. I’ll leave this open for now as the design discussion anchor unless you prefer a fresh PR.

kilhyeonjun:
Thanks for keeping this open and for the context. I rechecked the PR against current `main`; it now conflicts in `kiro/routes_anthropic.py`, so I agree it should not be merged as-is.

My current take on the architecture:

- For Anthropic-compatible clients such as Claude Code, the response `usage.input_tokens` should stay tied to the logical Anthropic request shape when we can compute or request that value.
- Kiro's `contextUsagePercentage` is still valuable, but it is post-facto utilization of the Kiro context window, not the same thing as Anthropic's upfront token count contract.
- Streaming and non-streaming should share the same accounting rule: use provider-reported/logical Anthropic count where available, and only fall back to the local estimate when Kiro does not return enough information.
- I would avoid letting the OpenAI route regress to local token math if Kiro already gives a better post-facto value there; that should stay a separate path from the Anthropic compatibility fix.

Given the conflict and your newer `count_tokens` implementation, I think the cleanest next step is for me to rebuild this on top of current `main` as a narrower patch/review, rather than forcing this branch through a rebase. If you prefer, I can also close this PR and open a fresh one with the streaming/non-streaming behavior documented in tests.

PR #149, opened 2026-05-02
fix(anthropic,openai): include tool_use content in output token calculation

tangyu:
## Summary

- `output_tokens` / `completion_tokens` calculation was ignoring tool_use block content (tool name + input arguments JSON), only counting text and thinking content
- A `write_file` tool call with ~4500 chars of content was reporting only 26 output_tokens instead of ~1300+
- Fixed all three code paths:
  - **Anthropic streaming** (`stream_kiro_to_anthropic`): includes `tool_blocks` content in `count_tokens()`
  - **Anthropic non-streaming** (`collect_anthropic_response`): includes `result.tool_calls` content in `count_tokens()`
  - **OpenAI streaming** (`stream_kiro_to_openai`): includes `all_tool_calls` content in `count_tokens()`. The non-streaming path (`collect_stream_response`) reads usage from streaming output, so it's automatically fixed.

## Root Cause

In all three paths, output tokens were calculated as:

```python
output_tokens = count_tokens(full_content + full_thinking_content)
```

When the response contains `tool_use` blocks, the tool name and input arguments were never fed into `count_tokens()`. For tool-only responses (no text), `full_content` is empty, resulting in near-zero output token counts.

## Test Results

| Protocol | Mode | Before | After |
|----------|------|--------|-------|
| Anthropic | streaming | 26 tokens | ~1334 tokens |
| Anthropic | non-streaming | 26 tokens | ~1334 tokens |
| OpenAI | streaming | ~26 tokens | ~2147 tokens |
| OpenAI | non-streaming | ~26 tokens | ~2158 tokens |

(All tests used a `write_file` tool call generating a Flask CRUD app, ~4500-6000 chars of tool input)

## Test plan

- [x] Anthropic streaming: tool_use response token count verified
- [x] Anthropic non-streaming: tool_use response token count verified
- [x] OpenAI streaming: tool_use response token count verified
- [x] OpenAI non-streaming: tool_use response token count verified
- [x] Text-only responses unaffected (no tool_blocks = no change)

### DIFF
--- a/kiro/streaming_anthropic.py
+++ b/kiro/streaming_anthropic.py
@@ -621,8 +621,12 @@ async def stream_kiro_to_anthropic(
                 f"{'Model will be notified automatically about truncation.' if TRUNCATION_RECOVERY else 'Set TRUNCATION_RECOVERY=true in .env to auto-notify model about truncation.'}"
             )
         
-        # Calculate output tokens
-        output_tokens = count_tokens(full_content + full_thinking_content)
+        # Calculate output tokens (include tool_use block content)
+        tool_content = ""
+        for tb in tool_blocks:
+            tool_content += json.dumps(tb.get("input", {}), ensure_ascii=False)
+            tool_content += tb.get("name", "")
+        output_tokens = count_tokens(full_content + full_thinking_content + tool_content)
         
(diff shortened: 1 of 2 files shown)


tangyu:

PR #150, opened 2026-05-02
fix: send profileArn for SSO OIDC users with explicit PROFILE_ARN config

ahmadumar020:
## Summary

- Users authenticating via AWS IAM Identity Center (SSO OIDC) with a Kiro Pro+ subscription set up through the AWS Console need `profileArn` in the request payload for usage to be tracked in the Kiro dashboard
- Previously `profileArn` was intentionally skipped for all SSO OIDC auth types (correct for Builder ID / free tier users, but wrong for Pro+ IAM Identity Center users)
- This fix sends `PROFILE_ARN` from `.env` when explicitly set, regardless of auth type

## Why this is safe

- Completely opt-in - only fires when `PROFILE_ARN` is explicitly set in `.env`
- Users who don't set `PROFILE_ARN` are unaffected - the `elif PROFILE_ARN:` branch never executes when the value is empty (the default)
- Existing Builder ID / free tier SSO OIDC users see no change in behaviour

## Use case

AWS Console -> Kiro -> Pro+ subscription with IAM Identity Center auth. The `profileArn` is required to attribute usage to the Pro+ account. Without it, requests go through as untracked free tier usage and the dashboard shows no activity.

## Test plan

- Set `PROFILE_ARN` in `.env` and verify `profileArn` appears in `debug_logs/kiro_request_body.json`
- Leave `PROFILE_ARN` unset and verify existing behaviour is unchanged
- Confirm usage appears in Kiro dashboard after setting `PROFILE_ARN`

### DIFF
--- a/kiro/routes_anthropic.py
+++ b/kiro/routes_anthropic.py
@@ -34,7 +34,7 @@ from fastapi.responses import JSONResponse, StreamingResponse
 from fastapi.security import APIKeyHeader
 from loguru import logger
 
-from kiro.config import PROXY_API_KEY
+from kiro.config import PROXY_API_KEY, PROFILE_ARN
 from kiro.models_anthropic import (
     AnthropicMessagesRequest,
     AnthropicCountTokensRequest,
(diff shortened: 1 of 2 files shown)


ahmadumar020:

ahmadumar020:

ahmadumar020:
@cla-bot check

ahmadumar020:
@cla-bot check

ahmadumar020:

ahmadumar020:
@cla-bot check

PR #156, opened 2026-05-10
feat(api): expose upstream tokenLimits on /v1/models

sbe27:
## Summary

Kiro's `ListAvailableModels` response carries per-model `tokenLimits`
(`maxInputTokens` / `maxOutputTokens`) which we already cache in
`ModelInfoCache` but strip before returning from `/v1/models`.

OpenAI-compatible clients probing the endpoint to size context windows
therefore fall back to conservative defaults (often 128K or 256K) even
for models that actually support 1M input (e.g. `claude-opus-4.6`,
`claude-opus-4.7`, `claude-sonnet-4.6`).

This PR surfaces those limits as three optional fields on each model
entry in the `/v1/models` response:

| Field | Maps to | Notes |
|---|---|---|
| `context_length` | `maxInputTokens` | Alias — the field name most OpenAI-compatible clients probe (LiteLLM, aider, Hermes, etc.) |
| `max_input_tokens` | `maxInputTokens` | Explicit form |
| `max_output_tokens` | `maxOutputTokens` | Explicit form |

All three default to `None` and are purely additive — stock OpenAI
clients that ignore unknown fields are unaffected.

## Example

Before:
```json
{
  "id": "claude-opus-4.7",
  "object": "model",
  "owned_by": "anthropic",
  "description": "Claude model via Kiro API"
}
```

After:
```json
{
  "id": "claude-opus-4.7",
  "object": "model",
  "owned_by": "anthropic",
  "description": "Claude model via Kiro API",
  "context_length": 1000000,
  "max_input_tokens": 1000000,
  "max_output_tokens": 64000
}
```

## Design notes

- **Highest-wins across accounts.** When the account system is enabled
  and the same model ID appears across multiple initialized accounts
  (different tiers), the response reports the *largest* limit observed.
  Callers want the achievable ceiling, not the strictest common
  denominator.

- **No private-attribute access.** Added two small public accessors:
  - `ModelInfoCache.get_token_limits()` — returns the raw
    `maxInputTokens` / `maxOutputTokens` dict (or `None`), letting
    callers distinguish "unknown" from the `DEFAULT_MAX_INPUT_TOKENS`
    fallback that `get_max_input_tokens()` already performs.
  - `AccountManager.iter_initialized_accounts()` — public read-only
    iterator over accounts with an initialized `auth_manager`.

- **Defensive wrapping.** The merge loop in `/v1/models` is wrapped in
  a `try/except` so a cache glitch can never 500 the endpoint.

## Tests

- `tests/unit/test_cache.py`: 4 new tests for `get_token_limits`
  covering happy path, unknown model, missing `tokenLimits`, and
  non-int values.
- `tests/unit/test_routes_openai.py`: 1 integration test that stands
  up a mocked `account_manager` with a known `tokenLimits` payload
  and verifies the new fields appear on the `/v1/models` response.

All 12 new/affected tests pass locally; 1,671 existing tests pass
(one pre-existing failure in `test_config.py::TestServerHostConfig`
caused by local `.env`, unrelated to this change).

## Verified live

Against a running 2.4-dev.10 gateway proxying real Kiro Pro accounts:

```
claude-haiku-4.5       ctx=200000  in=200000   out=64000
claude-opus-4.5        ctx=200000  in=200000   out=64000
claude-opus-4.6        ctx=1000000 in=1000000  out=64000
claude-opus-4.7        ctx=1000000 in=1000000  out=64000
claude-sonnet-4.5      ctx=200000  in=200000   out=64000
claude-sonnet-4.6      ctx=1000000 in=1000000  out=64000
deepseek-3.2           ctx=164000  in=164000   out=64000
qwen3-coder-next       ctx=256000  in=256000   out=64000
```

### DIFF
--- a/kiro/account_manager.py
+++ b/kiro/account_manager.py
@@ -808,6 +808,21 @@ class AccountManager:
             if account.auth_manager is not None:
                 return account
         raise RuntimeError("No initialized accounts available")
+
+    def iter_initialized_accounts(self):
+        """
+        Yield every account whose ``auth_manager`` has been initialized.
+
+        Public accessor intended for read-only introspection (e.g. building
+        the ``/v1/models`` response). Use this instead of reaching into
+        ``self._accounts`` directly.
+
+        Yields:
+            Account: initialized accounts, in insertion order.
+        """
+        for account in self._accounts.values():
+            if account.auth_manager is not None:
+                yield account
     
     def get_all_available_models(self) -> List[str]:
         """
--- a/kiro/cache.py
+++ b/kiro/cache.py
@@ -141,6 +141,35 @@ class ModelInfoCache:
             return model["tokenLimits"].get("maxInputTokens") or DEFAULT_MAX_INPUT_TOKENS
         return DEFAULT_MAX_INPUT_TOKENS
     
+    def get_token_limits(self, model_id: str) -> Optional[Dict[str, int]]:
+        """
+        Returns raw tokenLimits for the model, as reported by Kiro's
+        ListAvailableModels (``{"maxInputTokens": int, "maxOutputTokens": int}``).
+
+        Unlike :meth:`get_max_input_tokens`, this does not fall back to the
+        ``DEFAULT_MAX_INPUT_TOKENS`` constant — callers that want the *real*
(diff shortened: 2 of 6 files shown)


sbe27:

PR #158, opened 2026-05-11
fix: resolve file write failures for Claude Code and OpenCode via gateway

### LINKED ISSUE #56: Kiro API truncates large tool call payloads mid-stream
## Summary

When Claude generates tool calls with large payloads (e.g., Write tool with 200+ lines of code), the Kiro API truncates the response stream before the tool call completes. This results in incomplete JSON that cannot be parsed, causing tool calls to fail silently with empty parameters.

**This is an upstream Kiro API limitation, not a gateway bug.** The gateway correctly logs and handles the truncated data, but cannot recover the missing payload.

## Environment

- Kiro Gateway version: 2.1
- Client: Claude Code CLI
- Models affected: claude-opus-4-5, claude-sonnet-4-5 (likely all models)

## Symptoms

| Symptom | Description |
|---------|-------------|
| `Write: {}` in client transcript | Tool call recorded with empty parameters |
| `tool_blocks=0` in gateway logs | No tool calls detected despite model generating them |
| No `contextUsagePercentage` | Stream cut before completion metadata sent |
| No `"stop":true` event | Tool call never finalized |

### Truncated vs Working Session Comparison

| Metric | Truncated Session | Working Session |
|--------|-------------------|-----------------|
| Stream size | 4,775 bytes | 39,318 bytes |
| toolUseEvent count | 16 | 110 |
| `"stop":true` events | **0** | 2 |
| contextUsagePercentage | **Not received** | 61.49% |

### Raw Stream Analysis

The truncated stream shows the tool call starting correctly but being cut off mid-payload:

```
1. <thinking>...</thinking> - Completes normally
2. toolUseEvent: {"name":"Write","toolUseId":"..."} - Starts correctly
3. toolUseEvent: {"input":"{\"file_pa",...} - Building file_path incrementally
4. ... (16 chunks total)
5. toolUseEvent: {"input":"or.ts\"",...} - STREAM ENDS ABRUPTLY

Missing:
- The "content" parameter (hundreds of lines of code)
- Closing braces for the JSON
- The "stop":true event to finalize the tool call
- The contextUsagePercentage event
```

### Key Finding: No `max_tokens` Support

The Kiro API does not accept `max_tokens` or `inferenceConfig` parameters. The gateway payload only contains:
- `conversationState` (messages, history, tools)
- `profileArn`

This means **there is no client-side control over output limits** - truncation is entirely server-side.

## Reproduction Steps

1. Start a Claude Code session via kiro-gateway
2. Ask Claude to write a large file (200+ lines of code)
3. Enable `DEBUG_MODE=all` to capture raw stream
4. Observe the stream truncation in `debug_logs/response_stream_raw.txt`
5. Tool call fails with empty parameters

## Why This Is NOT a Gateway Bug

1. **Raw stream logging proves data stops arriving** - The gateway logs every byte received from Kiro API
2. **No timeout issues** - `STREAMING_READ_TIMEOUT=300s` is generous; stream ends cleanly at packet boundary
3. **Parser handles truncation correctly** - `_diagnose_json_truncation()` detects and logs the issue
4. **Working sessions complete normally** - Same code path works for smaller payloads

## Workarounds

Users can mitigate this by:

1. **Write smaller files** - Break large files into multiple smaller writes (<100 lines each)
2. **Use incremental edits** - Use Edit tool instead of Write for modifications
3. **Reduce context** - Start fresh sessions to minimize conversation history
4. **Manual copy** - Have Claude output code in chat, copy manually

## Questions for Maintainer

1. Is there a known output size limit in the Kiro API?
2. Are there any undocumented parameters to control response size?
3. Has this been reported to AWS/Amazon Q team?

## Suggested Documentation

Consider adding a note to the README about this limitation so users are aware and can use the workarounds proactively.

---

*Investigated with DEBUG_MODE=all. Raw stream evidence available if needed.*

lethanhson9901:
## Problem

Three root causes caused file write operations (`Write`, `Edit`) to silently fail when using Claude Code or OpenCode via kiro-gateway.

### Root Cause 1 — Tool argument corruption (`parsers.py`)

Kiro API streams tool call arguments as: `tool_start` (empty `input={}`) → `tool_input` fragments → `tool_stop`.

**Bug:** Gateway serialized the empty dict to `"{}"`, then appended fragments → `"{}{\"path\"...}"` (invalid JSON). `_finalize_tool_call()` fell back to `{}`, so the model received no `file_path`/`content` and the write silently failed.

**Fix:** Empty dict `{}` in `tool_start` → `""` so fragments concatenate cleanly.

### Root Cause 2 — Thinking tags injected into tool_result messages (`converters_core.py`)

When `FAKE_REASONING_ENABLED=True`, thinking tags were injected into every user message, including Turn 2 of a tool flow (which carries `toolResults`). Kiro API returned `"Improperly formed request"` when a message had both `toolResults` and thinking tags.

**Fix:** Skip thinking tag injection when the current message contains `tool_results`.

### Root Cause 3 — Kiro API ~9KB output limit on tool call streams

Kiro API silently truncates tool call output streams exceeding ~9KB. For large file writes, the stream is cut mid-JSON without a `tool_stop` event, causing fallback to `{}`.

**Fix:** Inject a system prompt addition (when tools are present) instructing the model to split large file writes into chunks using `Edit`/`edit`, keeping each call under ~7KB. Controlled by `TOOL_CALL_SIZE_GUARD` env var (default: `true`).

## Changes

| File | Change |
|------|--------|
| `kiro/parsers.py` | Fix empty dict `{}` → `""` in `tool_start` to prevent fragment corruption |
| `kiro/converters_core.py` | Skip thinking tags when message has `tool_results`; add `get_tool_call_size_guard_system_addition()` |
| `kiro/config.py` | Add `TOOL_CALL_SIZE_GUARD` constant (default: `true`) |
| `.env.example` | Document `TOOL_CALL_SIZE_GUARD` |

## Tests

- `TestToolStartEmptyDictFix` (6 tests) — `test_parsers.py`
- `TestThinkingTagsNotInjectedWithToolResults` (3 tests) — `test_converters_core.py`
- `TestNoThinkingTagsInToolResultMessages` (3 tests) — `test_converters_openai.py`
- `TestToolCallSizeGuard` (5 tests) — `test_converters_core.py`
- `tests/integration/test_file_write_flow.py` (12 end-to-end simulation tests)

## Verification

End-to-end tested with Claude Code writing files of 12KB–16.6KB:
- Model automatically chunks large writes into ~2–5KB pieces using `Edit`
- All files produced are syntactically valid Python
- No truncation errors in gateway logs

## Configuration

```env
# Disable if you handle chunking at the client level
TOOL_CALL_SIZE_GUARD=true  # default
```

Closes #56 (partial — addresses tool call truncation for file writes)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -281,6 +281,13 @@ PROXY_API_KEY="my-super-secret-password-123"
 # Default: true (ENABLED by default for better out-of-the-box experience)
 # TRUNCATION_RECOVERY=true
 
+# Tool Call Size Guard (default: true)
+# Injects a system prompt instruction to prevent large file writes from being
+# truncated by Kiro API's ~9KB tool call output limit (causes silent {} failure).
+# Instructs the model to split large Write/write calls into chunks using Edit/edit.
+# Disable only if you handle chunking at the client level (e.g., custom CLAUDE.md).
+# TOOL_CALL_SIZE_GUARD=true
+
 # ===========================================
 # LOGGING
 # ===========================================
--- a/docs/superpowers/plans/2026-05-13-tool-call-size-guard.md
+++ b/docs/superpowers/plans/2026-05-13-tool-call-size-guard.md
@@ -1,0 +1,249 @@
+# Tool Call Size Guard Implementation Plan
+
+> **For agentic workers:** Use subagent-driven-development (recommended) or executing-plans skill.
+
+**Goal:** Inject a system prompt addition that instructs the model to split large file writes into chunks, preventing Kiro API's ~9KB tool call output truncation.  
+**Architecture:** New config constant + new function in `converters_core.py` following the existing `get_thinking_system_prompt_addition()` pattern. Injected in `build_kiro_payload()` only when tools are present.  
+**Tech Stack:** Python, pytest
+
+---
+
+### Task 1: Add config constant
+
+**Files:**
+- Modify: `kiro/config.py` (after TRUNCATION_RECOVERY block, ~line 329)
+
+- [ ] **Step 1: Add constant**
+
+In `kiro/config.py`, after the `TRUNCATION_RECOVERY` line, add:
+
+```python
+# ==================================================================================================
+# Tool Call Size Guard Settings
+# ==================================================================================================
+
+# Inject system prompt instruction to prevent tool call arguments exceeding
+# Kiro API's ~9KB output limit (causes silent truncation to empty {})
+# Default: true (enabled) — disable only if you handle chunking at the client level
+TOOL_CALL_SIZE_GUARD: bool = os.getenv("TOOL_CALL_SIZE_GUARD", "true").lower() in ("true", "1", "yes")
+```
+
+- [ ] **Step 2: Verify import works**
+```bash
+cd /root/project/kiro-gateway && python3 -c "from kiro.config import TOOL_CALL_SIZE_GUARD; print(TOOL_CALL_SIZE_GUARD)"
+```
+Expected: `True`
+
+---
+
+### Task 2: Add function + integration in converters_core.py
+
+**Files:**
+- Modify: `kiro/converters_core.py`
+
+- [ ] **Step 1: Add function after `get_truncation_recovery_system_addition()`**
+
+```python
+def get_tool_call_size_guard_system_addition(tools: Optional[List["UnifiedTool"]]) -> str:
+    """
+    Generate system prompt addition that prevents large tool call arguments.
+
+    Kiro API silently truncates tool call output streams exceeding ~9KB,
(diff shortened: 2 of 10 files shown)


jwadow (commented):
not merging this pr. the parser fix is already covered by our own minimal commit. the tool size guard goes against our transparency philosophy since the gateway shouldn't tell the model how to write code, and our truncation recovery system already handles that by notifying the model when it happens. also, the pr has some issues: it includes docs/ files that shouldn't be in the repo, creates a new test file instead of using existing ones, and the CLA isn't signed. the truncation issue is just a Kiro API limitation that's already handled by our recovery system

PR #162, opened 2026-05-13
fix(openai): accept x-api-key for model discovery

Yorick-Ryu:
## Summary

Anthropic-style channel model discovery can call OpenAI-compatible endpoints with `x-api-key` instead of `Authorization: Bearer ...`. This PR lets the OpenAI route auth helper accept both header styles.

## Changes

- Add `x-api-key` as an accepted auth header on OpenAI-compatible routes
- Keep existing `Authorization: Bearer` behavior unchanged
- Add coverage for direct auth helper validation and `/v1/models`

## Test plan

- [x] Added unit test for `verify_api_key` accepting `x-api-key`
- [x] Added route test for `GET /v1/models` with `x-api-key`

## CLA

### DIFF
--- a/kiro/routes_openai.py
+++ b/kiro/routes_openai.py
@@ -29,6 +29,8 @@ Contains all API endpoints:
 import json
 from datetime import datetime, timezone
 
+from typing import Optional
+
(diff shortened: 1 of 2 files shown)


Yorick-Ryu:

PR #163, opened 2026-05-13
fix(anthropic): handle document blocks and structured output

Yorick-Ryu:
## Summary

Hvoy-style Anthropic clients can send document content blocks and structured-output settings that need gateway-side compatibility before forwarding to Kiro. This PR keeps those requests valid and preserves the intended prompt constraints.

## Changes

- Accept Anthropic document content blocks and extract text from simple base64 PDF documents
- Avoid web_search auto-injection when document content is present
- Add prompt guidance for tool_choice requests
- Add output_config json_schema guidance and suppress thinking prompt additions for JSON schema output

## Test plan

- [x] Added converter test for extracting text from Anthropic PDF document blocks
- [x] Added integration test that document text reaches the Kiro payload
- [x] Added converter test for tool_choice prompt guidance
- [x] Added converter test for output_config json_schema guidance and thinking suppression

## CLA

### DIFF
--- a/kiro/converters_anthropic.py
+++ b/kiro/converters_anthropic.py
@@ -24,6 +24,9 @@ This module is an adapter layer that converts Anthropic-specific formats
 to the unified format used by converters_core.py.
 """
 
+import base64
+import re
+import zlib
 from typing import Any, Dict, List, Optional
 
 from loguru import logger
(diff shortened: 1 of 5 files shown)


Yorick-Ryu:

PR #164, opened 2026-05-13
fix(models): add redacted_thinking content block support

Yorick-Ryu:
## Summary

Claude and Anthropic clients may replay redacted_thinking content blocks in Messages API history. The gateway should validate those official blocks while keeping the opaque redacted payload out of prompt text conversion.

## Changes

- Add RedactedThinkingContentBlock for Anthropic redacted_thinking content
- Include redacted_thinking in the ContentBlock union
- Keep converter behavior ignoring non-text redacted thinking payload data
- Add model, converter, and route coverage for replayed redacted thinking blocks

## Test plan

- [x] Added model validation tests for RedactedThinkingContentBlock
- [x] Added ContentBlock union test for redacted_thinking blocks
- [x] Added converter test ensuring redacted_thinking data is ignored as prompt text
- [x] Added /v1/messages route validation test for replayed redacted_thinking content

## CLA

### DIFF
--- a/kiro/models_anthropic.py
+++ b/kiro/models_anthropic.py
@@ -65,6 +65,20 @@ class ThinkingContentBlock(BaseModel):
     signature: str = ""
 
 
+class RedactedThinkingContentBlock(BaseModel):
+    """
+    Redacted thinking content block in Anthropic format.
+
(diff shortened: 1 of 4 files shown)


Yorick-Ryu:

PR #165, opened 2026-05-14
[codex] Support Anthropic document PDF blocks

ajsh:
## Summary
- Accept Anthropic `document` content blocks in Messages API requests, including Claude Code PDF file-read results.
- Convert base64 PDF and text document blocks into plain text for the Kiro payload instead of forwarding raw base64.
- Add regression coverage for Claude Code's `tool_result` plus PDF `document` shape.

## Root Cause
Claude Code sends PDFs as Anthropic `document` blocks after reading a local PDF. Kiro Gateway's Anthropic schema only accepted text, thinking, image, tool use/result, and tool reference blocks, so those requests failed validation with HTTP 422 before conversion.

## Validation
- `python3 -m pytest tests/unit/test_models_anthropic.py tests/unit/test_converters_anthropic.py -q`

### DIFF
--- a/kiro/converters_anthropic.py
+++ b/kiro/converters_anthropic.py
@@ -24,6 +24,8 @@ This module is an adapter layer that converts Anthropic-specific formats
 to the unified format used by converters_core.py.
 """
 
+import base64
+from io import BytesIO
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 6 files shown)


ajsh:

PR #169, opened 2026-05-16
fix(schema): drop invalid required values

bunizao:
## Summary
- remove invalid `required` values from sanitized JSON schemas
- keep non-empty `required` arrays unchanged
- add regression coverage for null and non-array `required` values, including tool input schema conversion

## Root Cause
Bedrock rejects tool schemas when `inputSchema.json.required` is present but not an array. The gateway already removed empty arrays for Kiro compatibility, but `required: null` passed through unchanged and triggered `TOOL_SCHEMA_INVALID`.

## Validation
- `.venv/bin/python -m pytest tests/unit/test_converters_core.py::TestSanitizeJsonSchema tests/unit/test_converters_core.py::TestConvertToolsToKiroFormat -q`
- `.venv/bin/python -m pytest tests/unit/test_converters_core.py -q`

Note: `.venv/bin/python -m pytest tests/unit -q` currently reaches 1598 passing tests and then fails 79 route tests during TestClient startup because this local checkout has no `credentials.json`; those failures are unrelated to this schema sanitizer change.

### DIFF
--- a/kiro/converters_core.py
+++ b/kiro/converters_core.py
@@ -441,7 +441,7 @@ def sanitize_json_schema(schema: Optional[Dict[str, Any]]) -> Dict[str, Any]:
     Sanitizes JSON Schema from fields that Kiro API doesn't accept.
     
     Kiro API returns 400 "Improperly formed request" error if:
-    - required is an empty array []
+    - required is not a non-empty array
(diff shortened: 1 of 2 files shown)


bunizao:

bunizao:

PR #179, opened 2026-05-20
feat(codex): Add /v1/responses API endpoint

nicandcranny:
## Summary
- add OpenAI-compatible POST /v1/responses for Codex Desktop
- translate Responses API input, instructions, function calls, function outputs, and flat tools into the existing Kiro chat conversion path
- emit Responses API response objects and streaming SSE events, including text deltas and function-call argument events
- normalize credential paths so configured ~ paths are expanded consistently

## Why
Codex Desktop uses the OpenAI Responses API surface rather than Chat Completions. This lets Kiro Gateway remain transparent while accepting that newer client API shape and reusing the existing Kiro conversion, streaming, auth, and retry behavior.

## Tests
- venv/bin/python -m pytest -v
- Tested from Codex desktop app with this configuration in `~/.codex/config.toml`

```
model = "claude-opus-4-6"
model_provider = "kiro"

[model_providers.kiro]
name = "Kiro Gateway"
base_url = "http://localhost:9999/v1"
env_key = "KIRO_GATEWAY_API_KEY"
wire_api = "responses"
```

image

### DIFF
--- a/kiro/__init__.py
+++ b/kiro/__init__.py
@@ -61,6 +61,7 @@ from kiro.config import (
 # Models
 from kiro.models_openai import (
     ChatCompletionRequest,
+    ResponsesRequest,
     ChatMessage,
     OpenAIModel,
     ModelList,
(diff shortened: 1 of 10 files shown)


nicandcranny:

nicandcranny:
> Thanks for the PR! 🎉
> 
> 
> 
> Please reply once with:
> 
> ```
> ```
> 


nicandcranny:

PR #182, opened 2026-05-21
fix: truncate long MCP tool names instead of rejecting requests

glyphx:
## Problem

Claude Code's MCP plugin naming convention uses the format `mcp__<server>_<plugin>__<method>` which can easily exceed the Kiro API's 64-character tool name limit. For example:

```
mcp__plugin_chrome-devtools-mcp_chrome-devtools__get_console_message (68 chars)
```

The gateway currently rejects these requests with a 400 error:
```
API Error: 400 Tool name(s) exceed Kiro API limit of 64 characters:
  - 'mcp__plugin_chrome-devtools-mcp_chrome-devtools__get_console_message' (68 characters)
```

This makes it impossible to use many MCP plugins with Claude Code through the gateway.

## Fix

Instead of rejecting requests, automatically truncate long tool names to fit within 64 characters using a deterministic scheme:
- Keep the first 56 characters (preserves readability)
- Append `_` + 7-character MD5 hash (ensures uniqueness)

A bidirectional mapping is maintained so that tool names in responses from the Kiro API are restored to their original values before being sent back to Claude Code. This ensures the client never sees the truncated names.

### Example
```
Original:  mcp__plugin_chrome-devtools-mcp_chrome-devtools__get_console_message (68 chars)
Truncated: mcp__plugin_chrome-devtools-mcp_chrome-devtools__get_con_8d29446 (64 chars)
Response:  restored to original before reaching Claude Code
```

## Files Changed

- **converters_core.py:** Replace `validate_tool_names` (which raised ValueError) with truncation + reverse mapping logic
- **streaming_anthropic.py:** Apply `get_original_tool_name()` to restore names in Anthropic-format responses
- **streaming_openai.py:** Same for OpenAI-compatible response path

## Testing

Installed Chrome DevTools MCP plugin in Claude Code — requests now pass through the gateway successfully instead of being rejected.

### DIFF
--- a/kiro/converters_core.py
+++ b/kiro/converters_core.py
@@ -557,46 +557,45 @@ def process_tools_with_long_descriptions(
     return processed_tools if processed_tools else None, tool_documentation
 
 
+import hashlib
+
+# Global mapping for tool name truncation (short_name -> original_name)
+_tool_name_map: Dict[str, str] = {}
+
+
+def _shorten_tool_name(name: str) -> str:
+    """Shorten a tool name to fit the 64-char Kiro API limit using a hash suffix."""
+    if len(name) <= 64:
+        return name
+    # Keep first 56 chars + underscore + 7-char hash for uniqueness
+    h = hashlib.md5(name.encode()).hexdigest()[:7]
+    return name[:56] + "_" + h
+
+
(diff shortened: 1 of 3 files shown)


glyphx:

glyphx:

PR #183, opened 2026-05-21
fix(tokenizer): cache cl100k_base locally and retry on network errors

1c3z:
`_get_encoding` used to mark tiktoken unusable on ANY exception, including a transient `IncompleteRead` while pulling cl100k_base.tiktoken from the Azure blob CDN. A single dropped connection on first use permanently demoted the whole process to the length-based fallback heuristic, silently losing tokenization precision for every later request until restart.

Two changes:

1. Pin `TIKTOKEN_CACHE_DIR` (via `setdefault`, so operator overrides win) to a repo-local `.tiktoken_cache/` directory. The ~1.6 MB BPE blob is fetched at most once per working copy and reused across restarts, eliminating the per-process re-download that exposed us to CDN flakiness in the first place. `.tiktoken_cache/` added to .gitignore.

2. Separate `ImportError` (tiktoken truly missing → cache the disabled state permanently) from runtime errors (network/IO failure during BPE download → do NOT cache, return None this call, retry on next call). One transient blip no longer kills tokenization for the lifetime of the process.

Tests (tests/unit/test_tokenizer.py::TestEncodingCacheAndRetry):
- TIKTOKEN_CACHE_DIR is set to repo-local `.tiktoken_cache` on module import.
- Network failure does not poison the `_encoding` sentinel.
- After a failure, the next call retries and a recovered encoding is cached.
- `ImportError` is still cached so we don't re-import a missing package on every call (regression guard for the previous behavior).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ build/
 debug_logs*/
 debug*.json
 
+# tiktoken BPE cache (downloaded on first use, see kiro/tokenizer.py)
+.tiktoken_cache/
+
 # Project-specific
 _notes/
 requests/
--- a/kiro/tokenizer.py
+++ b/kiro/tokenizer.py
@@ -33,11 +33,29 @@ more than GPT-4 (cl100k_base). This is due to differences in BPE vocabularies.
 """
 
 import json
+import os
 from typing import List, Dict, Any, Optional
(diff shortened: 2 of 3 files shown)


1c3z:

PR #184, opened 2026-05-21
fix(model_resolver): resolve MODEL_ALIASES in get_model_id_for_kiro

emreyn1:
`MODEL_ALIASES` weren't being resolved in `get_model_id_for_kiro()`, which is the path the converters use in legacy mode (`ACCOUNT_SYSTEM=false`). Custom aliases like `kiro-sonnet` got sent to the Kiro API as-is and were rejected with "Invalid model ID" instead of being resolved to e.g. `claude-sonnet-4.5`. Account-system mode resolved them correctly, only the legacy converter path was skipping the lookup.

The use case where this hurts: pointing Cursor IDE's "custom OpenAI base URL" at the gateway. `claude-*` names conflict with Cursor's built-in model names, so the only way to use a Kiro Claude model is via `MODEL_ALIASES`, and that wasn't working.

Fix: alias resolution is now the first step in `get_model_id_for_kiro()`, so the chain is `alias → normalize → hidden lookup → runtime ID`.

Added 5 tests in `TestGetModelIdForKiro` covering aliased / non-aliased / alias-to-hidden / empty-alias-dict cases:

- `test_resolves_alias_before_normalization`
- `test_alias_resolution_then_normalization`
- `test_alias_to_hidden_model`
- `test_non_aliased_model_unchanged`
- `test_empty_aliases_dict`

All 123 tests in `test_model_resolver.py` pass; verified end-to-end against Cursor IDE over a Cloudflare tunnel.

### DIFF
--- a/kiro/model_resolver.py
+++ b/kiro/model_resolver.py
@@ -192,20 +192,20 @@ def normalize_model_name(name: str) -> str:
 def get_model_id_for_kiro(model_name: str, hidden_models: Dict[str, str]) -> str:
     """
     Get the model ID to send to Kiro API.
-    
+
     This is a simple helper for converters that don't have access to the full
-    ModelResolver. It normalizes the name and checks hidden models.
-    
+    ModelResolver. It resolves aliases, normalizes the name, and checks hidden models.
+
     For hidden models (like claude-3.7-sonnet), returns the internal Kiro ID.
     For regular models, returns the normalized name.
-    
+
     Args:
         model_name: External model name from client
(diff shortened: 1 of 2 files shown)


emreyn1:

emreyn1:

emreyn1:
@jwadow I've accepted the CLA terms above — could you add `emreyn1` to the `.clabot` contributors list so the check passes? Thanks!

emreyn1:
hello @jwadow when you are free could you confirm the pr

PR #188, opened 2026-05-25
feat(routes_openai): support x-api-key header for /v1/models

gaoshilei:
## Summary

The Anthropic-shaped routes (`routes_anthropic.py`) accept **both** `x-api-key`
and `Authorization: Bearer` for compatibility, but the OpenAI-shaped routes
(`routes_openai.py`) only accept `Authorization: Bearer`. This causes
authentication mismatches for clients that auto-pick a single auth header
based on the API mode they're using.

This PR adds `x-api-key` as a fallback auth method on the OpenAI routes,
mirroring the dual-auth scheme already used by the Anthropic routes. Bearer
remains the primary mechanism — existing clients are unaffected.

## Motivation

I hit this while integrating [Hermes Agent](https://github.com/NousResearch/hermes-agent)
(an open-source AI agent framework by Nous Research) with kiro-gateway as a custom
provider. Hermes implements per-provider `api_mode` and uses **a single auth
header for all endpoints** of that provider:

```python
# hermes-agent/hermes_cli/models.py — probe_api_models()
if api_key and api_mode == "anthropic_messages":
    headers["x-api-key"] = api_key
    headers["anthropic-version"] = "2023-06-01"
elif api_key:
    headers["Authorization"] = f"Bearer {api_key}"
```

When configured with `api_mode: anthropic_messages` (so that `/v1/messages`
works), hermes also sends `x-api-key` to `/v1/models` for the model picker
(`hermes -p <profile> model`). Today this returns:

```
$ hermes -p yuanshen model
  Provider: kiro-gateway
  URL:      http://127.0.0.1:8788
  Current:  claude-sonnet-4.6

Fetching available models...
Could not fetch models from endpoint.
Model name [claude-sonnet-4.6]:
```

Gateway logs show two requests fired in parallel — the first auth attempt
gets 401 because hermes uses `x-api-key`, then a second attempt (with a
different code path) succeeds with `Authorization: Bearer`. By the time the
second one returns, the picker has already given up:

```
127.0.0.1:52352 - "GET /v1/models HTTP/1.1" 401   ← x-api-key, rejected
127.0.0.1:52353 - "GET /v1/models HTTP/1.1" 200   ← Bearer, accepted (too late)
```

This is a real interop issue: any client built around the assumption that the
same `api_mode` should produce the same auth header for both `/v1/models` and
`/v1/messages` will hit it.

### `kiro/routes_openai.py`

`verify_api_key` now accepts both headers and tries them in order:

1. `Authorization: Bearer {PROXY_API_KEY}` (OpenAI native, primary — existing
   behavior preserved exactly)
2. `x-api-key: {PROXY_API_KEY}` (Anthropic native, fallback — new)
3. Otherwise → 401

```python
async def verify_api_key(
    auth_header: str = Security(api_key_header),
    x_api_key: str = Security(x_api_key_header),
) -> bool:
    # Check Authorization: Bearer first (OpenAI native)
    if auth_header and auth_header == f"Bearer {PROXY_API_KEY}":
        return True

    # Fall back to x-api-key (Anthropic native)
    if x_api_key and x_api_key == PROXY_API_KEY:
        return True

    logger.warning("Access attempt with invalid API key.")
    raise HTTPException(status_code=401, detail="Invalid or missing API Key")
```

This keeps the existing Bearer flow as the primary path. The order
intentionally **differs** from `routes_anthropic.verify_anthropic_api_key`
(which checks `x-api-key` first) — each route checks its own native header
first to keep the prevalent client behavior on its fast path.

### Tests

Added 5 new test cases in `TestVerifyApiKey` (mirroring the dual-auth tests
already present for `verify_anthropic_api_key`):

- `test_valid_x_api_key_returns_true` — happy path for the new header
- `test_invalid_x_api_key_raises_401` — wrong key via x-api-key still rejected
- `test_bearer_takes_precedence_over_x_api_key` — Bearer wins when both are valid+invalid
- `test_x_api_key_used_when_bearer_invalid` — fallback works
- `test_missing_both_headers_raises_401` — defense in depth

The 7 existing tests were updated from positional to keyword args
(`verify_api_key(auth_header=..., x_api_key=...)`) for clarity given the
expanded signature; behavior unchanged.

```
$ pytest tests/unit/test_routes_openai.py -v
============================= 82 passed in 0.23s ==============================

$ pytest tests/unit/test_routes_openai.py::TestVerifyApiKey -v
12 passed (6 existing + 5 new + 1 already-existing empty-key test)
```

## Backwards Compatibility

- ✅ Clients using `Authorization: Bearer` (curl, OpenAI SDK, Cline,
  Continue, LangChain, etc.) are **completely unaffected** — the Bearer
  check is still first and runs before x-api-key
- ✅ No changes to `/v1/messages` (already supported both)
- ✅ No new dependencies; only standard FastAPI `APIKeyHeader`
- ✅ Public function signature changes from
  `verify_api_key(auth_header)` → `verify_api_key(auth_header, x_api_key=None)`,
  but `x_api_key` has `Security(...)` default so it's only positional-callable
  for tests, and even those are fine because it has a default value

## Verification

Tested against the Hermes Agent case end-to-end:

```bash
# Before this PR
$ hermes -p yuanshen model
Fetching available models...
Could not fetch models from endpoint.
Model name [claude-sonnet-4.6]:    # falls back to manual entry, picker dead

# After this PR
$ hermes -p yuanshen model
Fetching available models...
Found 13 model(s):
Select model from kiro-gateway:
  -> claude-sonnet-4.6 (current)
     claude-opus-4.7
     claude-opus-4.6
     claude-haiku-4.5
     ...
     Cancel
```

curl reproduction:

```bash
KEY="$PROXY_API_KEY"

# All three return 200 after this PR (only the Bearer one returned 200 before):
curl -sS -o /dev/null -w "%{http_code}\n" http://localhost:8000/v1/models \
  -H "Authorization: Bearer $KEY"            # 200 (unchanged)
curl -sS -o /dev/null -w "%{http_code}\n" http://localhost:8000/v1/models \
  -H "x-api-key: $KEY"                       # 200 (new — was 401)
curl -sS -o /dev/null -w "%{http_code}\n" http://localhost:8000/v1/models  # 401 (security check intact)
```

## Related

- CONTRIBUTING.md emphasizes **Complete consistency**:
  > "Changes must be applied to BOTH OpenAI and Anthropic APIs"

  This PR brings the OpenAI route's auth scheme up to parity with the
  Anthropic route's, which already has dual-auth.

- No linked issue — discovered while building local infrastructure on top of
  kiro-gateway. Happy to open a tracking issue first if you'd prefer.

---

## Checklist (per CONTRIBUTING.md)

- [x] English-only code, comments, and docstrings
- [x] Type hints on the new signature (`auth_header: str`, `x_api_key: str`)
- [x] Google-style docstring on `verify_api_key` (Args / Returns / Raises)
- [x] `loguru` warning preserved on auth failure
- [x] HTTPException uses identical detail message ("Invalid or missing API Key")
- [x] No placeholders or TODOs
- [x] Focused diff: 2 files, no whitespace/format churn elsewhere
- [x] Tests added: 5 new (happy path, invalid key, precedence, fallback, defense)
- [x] All 82 tests in `test_routes_openai.py` pass

---

## Maintainer notes

If this lands, the local fork-and-patch workflow my team uses (a single
`patches` branch rebased onto `main`) becomes empty and we can drop it.
Happy to also send a follow-up if you want this consistency check added to
`tests/integration/test_full_flow.py`.

### DIFF
--- a/kiro/routes_openai.py
+++ b/kiro/routes_openai.py
@@ -62,28 +62,46 @@ except ImportError:
 
 
 # --- Security scheme ---
+# Authorization: Bearer (OpenAI native)
 api_key_header = APIKeyHeader(name="Authorization", auto_error=False)
+# x-api-key (Anthropic native) — for compatibility with clients that auto-detect
+# auth header by api_mode (e.g. hermes-cli with api_mode: anthropic_messages
+# uses x-api-key for /v1/models too). See PR #TBD.
+x_api_key_header = APIKeyHeader(name="x-api-key", auto_error=False)
 
 
-async def verify_api_key(auth_header: str = Security(api_key_header)) -> bool:
+async def verify_api_key(
+    auth_header: str = Security(api_key_header),
+    x_api_key: str = Security(x_api_key_header),
+) -> bool:
     """
-    Verify API key in Authorization header.
-    
-    Expects format: "Bearer {PROXY_API_KEY}"
-    
+    Verify API key for OpenAI-compatible endpoints.
+
+    Supports two authentication methods:
+    1. Authorization: Bearer {PROXY_API_KEY} (OpenAI native)
+    2. x-api-key: {PROXY_API_KEY} (Anthropic native, for clients that
+       send the same header for both /v1/models and /v1/messages)
+
     Args:
         auth_header: Authorization header value
-    
+        x_api_key: x-api-key header value
+
     Returns:
         True if key is valid
-    
+
     Raises:
-        HTTPException: 401 if key is invalid or missing
+        HTTPException: 401 if neither header has a valid key
     """
-    if not auth_header or auth_header != f"Bearer {PROXY_API_KEY}":
-        logger.warning("Access attempt with invalid API key.")
-        raise HTTPException(status_code=401, detail="Invalid or missing API Key")
-    return True
+    # Check Authorization: Bearer first (OpenAI native)
+    if auth_header and auth_header == f"Bearer {PROXY_API_KEY}":
+        return True
+
+    # Fall back to x-api-key (Anthropic native)
+    if x_api_key and x_api_key == PROXY_API_KEY:
+        return True
+
+    logger.warning("Access attempt with invalid API key.")
+    raise HTTPException(status_code=401, detail="Invalid or missing API Key")
 
 
 # --- Router ---
--- a/tests/unit/test_routes_openai.py
+++ b/tests/unit/test_routes_openai.py
@@ -43,7 +43,7 @@ class TestVerifyApiKey:
         valid_header = f"Bearer {PROXY_API_KEY}"
         
         print("Action: Calling verify_api_key...")
-        result = await verify_api_key(valid_header)
+        result = await verify_api_key(auth_header=valid_header, x_api_key=None)
         
         print(f"Comparing result: Expected True, Got {result}")
         assert result is True
@@ -59,7 +59,7 @@ class TestVerifyApiKey:
         
         print("Action: Calling verify_api_key with invalid key...")
         with pytest.raises(HTTPException) as exc_info:
-            await verify_api_key(invalid_header)
+            await verify_api_key(auth_header=invalid_header, x_api_key=None)
         
         print(f"Checking: HTTPException with status 401...")
         assert exc_info.value.status_code == 401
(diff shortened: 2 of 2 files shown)


gaoshilei:

PR #189, opened 2026-05-25
fix(mcp): include profileArn for Enterprise accounts in web_search

ShiuanTHC:
MCP web_search calls were failing for Enterprise (KIRO_DESKTOP) accounts because the request body did not include profileArn. Other API paths (converters_core, account_manager) already pass profileArn at the top level of the request body; mcp_tools was the only caller missing it.

Add profileArn to the MCP request body when auth_manager.profile_arn is set, mirroring the existing pattern in converters_core.py.

Also expand the non-200 error log to include a truncated response body, which makes diagnosing API errors (auth, validation, etc.) tractable without enabling DEBUG_MODE.

Tests:
- Add test verifying profileArn is included at top level for Enterprise auth managers (and not nested inside params).
- Add test verifying profileArn is omitted when auth_manager has no profile ARN (free / non-Enterprise accounts).

### DIFF
--- a/kiro/mcp_tools.py
+++ b/kiro/mcp_tools.py
@@ -135,6 +135,11 @@ async def call_kiro_mcp_api(
             "arguments": {"query": query}
         }
     }
+
+    # Add profileArn if available (required for Enterprise accounts)
+    # profileArn goes at the top level of the request body, not inside params
+    if auth_manager.profile_arn:
+        mcp_request["profileArn"] = auth_manager.profile_arn
     
     # Log MCP request
     try:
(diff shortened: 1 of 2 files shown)


ShiuanTHC:

hnewcity:
```
    """
    Message in Anthropic format.

    Attributes:
        role: Message role. Anthropic spec only allows user/assistant, but newer
            Claude Code clients sometimes inline "system" messages in the array.
            We accept it here and let normalize_message_roles() in converters_core
            collapse it to "user" downstream.
        content: Message content (string or list of content blocks)
    """

    content: Union[str, List[ContentBlock]]

    model_config = {"extra": "allow"}

```

Add "system" as a role works for me.

PR #191, opened 2026-05-29
fix(models): accept inline non-standard message roles (#190)

### LINKED ISSUE #190: BUG: 422 error when newer Claude Code client inlines `role: "system"` in messages array
### Kiro Gateway Version

v2.4.0

### What happened?

  ## Summary

  ## Error
  ```json
  API Error: 422 {
    "detail": [{
      "type": "literal_error",
      "input": "system",
    }],
    ...
  }
  ```
  Request body excerpt:

  ```json
  {
    "model": "claude-opus-4-8",
    "messages": [
    ]
  }
  ```
  Steps to reproduce

  A minimal curl repro:
  ```
    -d '{
      "model": "claude-sonnet-4-5",
      "max_tokens": 16,
      "messages": [
      ]
    }'
  ```
  Root cause

  unknown roles into "user".

  Expected behavior

### Debug Logs

Image

Goserizz:
## Summary

Newer Claude Code clients (`claude-opus-4-8`) inline a message with `role: "system"` inside the `messages` array — typically at index 1, carrying context blobs like `<system-reminder>` and `currentDate`. The gateway rejected these requests with **HTTP 422** before any downstream normalization could run, so the client could not complete a single turn.

```json
{
  "type": "literal_error",
  "input": "system"
}
```

Fixes #190.

## Root cause

`AnthropicMessage.role` in `kiro/models_anthropic.py` was declared as `Literal["user", "assistant"]`, so Pydantic rejected `"system"` at the validation layer — *before* the request reached `converters_core.normalize_message_roles()`, which already knows how to fold unknown roles into `"user"`.

## Fix

Widen the `role` field to a free-form `str`. Unknown roles such as `"system"` and `"developer"` continue to be normalized to `"user"` downstream by `normalize_message_roles()`, so the request now flows through correctly instead of being rejected up front. This keeps the gateway lenient on input while the official Anthropic spec (user/assistant only) is still honored on output.

## Tests

Added `TestAnthropicMessageInlineRoles` covering the inline `"system"` role at both the message and full-request level, plus a guard that standard `user`/`assistant` roles are unaffected.

```
425 passed
```
(`tests/unit/test_models_anthropic.py`, `test_converters_anthropic.py`, `test_converters_core.py`)

### DIFF
--- a/kiro/models_anthropic.py
+++ b/kiro/models_anthropic.py
@@ -183,11 +183,14 @@ class AnthropicMessage(BaseModel):
     Message in Anthropic format.
 
     Attributes:
-        role: Message role (user or assistant)
+        role: Message role. Normally "user" or "assistant", but accepted as a
+            free-form string because some clients (e.g. Claude Code) send other
+            roles such as "system" inside the messages array. Unknown roles are
+            normalized to "user" downstream by normalize_message_roles().
         content: Message content (string or list of content blocks)
     """
 
-    role: Literal["user", "assistant"]
+    role: str
     content: Union[str, List[ContentBlock]]
 
     model_config = {"extra": "allow"}
--- a/tests/unit/test_models_anthropic.py
+++ b/tests/unit/test_models_anthropic.py
@@ -623,6 +623,81 @@ class TestAnthropicMessagesRequestWithImages:
 
 
 # ==================================================================================================
+# Tests for inline non-standard message roles (Issue #190)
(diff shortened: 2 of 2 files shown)


Goserizz:

Goserizz:
@cla-bot check

1679558348:
but , when i ask: who you are,  it says:  i am kiro ....  so  still some problem

PR #192, opened 2026-05-29
feat(thinking): support native adaptive reasoning blocks

hawkff:
## Summary

This PR adds opt-in support for Claude/Kiro native adaptive thinking while preserving the existing fake-reasoning behavior by default.

It also extends OpenAI-compatible `reasoning_effort` handling to accept `max`, so clients can request Claude’s maximum native adaptive thinking effort through the gateway.

## What Changed

- Added `reasoning_effort="max"` to the OpenAI-compatible request schema.
- Moved reasoning-effort budget mapping into shared converter logic.
- Added experimental native adaptive thinking config:
  - `KIRO_NATIVE_THINKING_MODE=off|auto|force`
  - `KIRO_NATIVE_THINKING_DISPLAY=summarized|omitted`
- When native mode is enabled for supported models, the gateway sends Kiro/Claude native thinking fields:
  - `thinking: { type: "adaptive", display: "summarized" }`
  - `output_config: { effort: "max" }`
- Native thinking is feature-flagged and defaults to `off`, preserving current behavior.
- Added parser support for Kiro `reasoningContentEvent` frames, including text and signature events.
- Streams native reasoning through existing OpenAI/Anthropic response paths:
  - OpenAI: `reasoning_content` in semantic mode.
  - Visible tagged content in `pass` mode for clients that do not render `reasoning_content`.
  - Anthropic: native-style thinking blocks through the existing thinking response path.
- Added parser state handling so native thinking sequences open/close correctly across signature boundaries.
- Updated `.gitignore` to keep local compose overrides and local agent skill files out of commits.

## Why

Claude Opus 4.8 supports adaptive thinking, but the gateway previously only exposed fake prompt-tag reasoning. That meant clients could request higher effort levels, but the request did not activate Kiro’s native reasoning stream.

This change keeps fake reasoning available, but adds an opt-in native path for models that support adaptive thinking.

## Compatibility

The native thinking mode is intentionally disabled by default:

```bash
KIRO_NATIVE_THINKING_MODE=off
```

To enable it only when clients explicitly request effort:

```bash
KIRO_NATIVE_THINKING_MODE=auto
KIRO_NATIVE_THINKING_DISPLAY=summarized
```

To force native thinking for supported models even without a client effort value:

```bash
KIRO_NATIVE_THINKING_MODE=force
```

Existing fake reasoning env vars continue to work.

## Validation

- `python -m pytest -q`
  - `1702 passed`
- Live Kiro/Gateway probe confirmed `reasoning_effort=max` sends:
  - `thinking.type=adaptive`
  - `thinking.display=summarized`
  - `output_config.effort=max`
- Live low-vs-max probe on the same prompt:
  - `low`: 65 reasoning events, 648 reasoning chars
  - `max`: 97 reasoning events, 1006 reasoning chars
- Local CodeRabbit CLI review was run against the upstream base commit.
  - It found one minor issue around native thinking close events with empty content.
  - The issue was fixed.
  - Follow-up local CodeRabbit review of the fix returned `findings: 0`.
- TruffleHog scan on the final commit range:
  - `verified_secrets: 0`
  - `unverified_secrets: 0`

## Notes

Kiro’s native thinking fields are not publicly documented for the private `generateAssistantResponse` endpoint. They were validated with live upstream probes against the runtime API. For that reason, this PR keeps the native path behind explicit environment flags.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,10 +4,19 @@
 credentials.json
 state.json
 
+# Local Docker Compose overrides
+docker-compose.yaml
+docker-compose.local.yml
+docker-compose.local.yaml
+compose.local.yml
+compose.local.yaml
+
 # IDE
 .vscode/
 .idea/
 .shard/
+.agents/
+skills-lock.json
 
 # Python
 __pycache__/
--- a/kiro/config.py
+++ b/kiro/config.py
@@ -479,6 +479,26 @@ FAKE_REASONING_INITIAL_BUFFER_SIZE: int = int(os.getenv("FAKE_REASONING_INITIAL_
 
 
 # ==================================================================================================
+# Native Thinking Settings (Kiro/Claude Adaptive Thinking)
+# ==================================================================================================
+
+# Experimental native thinking pass-through for Kiro models that expose Claude adaptive thinking.
+#
+# Modes:
+# - "off": Disabled (default, preserves existing fake reasoning behavior)
+# - "auto": Enable only when the client explicitly requests a reasoning effort
+# - "force": Enable for supported models even when the client does not send an effort
+KIRO_NATIVE_THINKING_MODE: str = os.getenv("KIRO_NATIVE_THINKING_MODE", "off").lower()
+if KIRO_NATIVE_THINKING_MODE not in ("off", "auto", "force"):
+    KIRO_NATIVE_THINKING_MODE = "off"
+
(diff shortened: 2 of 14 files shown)


hawkff:

hawkff:
Manual smoke test through OpenCode using the Kiro Gateway provider.

Claude Opus 4.8, reasoning effort `max`
OpenCode renders a visible `Thought:` block before the final answer, confirming the gateway is surfacing reasoning content.
thinking_blocks

Issue #64, opened 2026-02-03
BUG: API error calling from codex App


walleXD:
### Kiro Gateway Version

latest

### What happened?

Getting Improperly formed request while calling from Codex App:

```toml
[model_providers.kiro]
name = "Kiro"
base_url = "http://kiro-gw.kiro-proxy.orb.local/v1"
wire_api = "chat"
experimental_bearer_token = "kiro"
```

Debug Logs attached

### Debug Logs

[debug_logs.zip](https://github.com/user-attachments/files/25030932/debug_logs.zip)

jwadow:
Hey @walleXD, fixed the issue.
The gateway now handles the `developer` role properly by converting it to a user message internally. Your requests should work now without any "Improperly formed request" errors.

Give it a try and let me know if you run into any issues.

PR #194, opened 2026-05-30
fix(models): accept inline 'system' role in Anthropic messages

severity1:
## Summary
- Widens `AnthropicMessage.role` from `Literal["user", "assistant"]` to `Literal["user", "assistant", "system"]` so inline `system` messages reach the existing `normalize_message_roles` pipeline (#64) instead of 422'ing at Pydantic validation.
- Aligns the Anthropic input path with the OpenAI input path, which already accepts and normalizes `system` via the same shared core.
- Adds 4 model-level tests (`TestAnthropicMessageRole`) and 1 end-to-end converter test (`test_inline_system_role_is_normalized_to_user`).

## Why
Claude Code on certain model ids (observed reliably with `claude-opus-4-8`, including the `[1m]` context-window variant) serializes inline `system` messages instead of using the top-level `system` field. The strict `Literal["user", "assistant"]` rejected these at the boundary with HTTP 422 before the existing recovery logic in `converters_core.py:1204` could collapse the role to `user`.

The Anthropic spec defines `system` as a top-level field, not a valid message role — so the strict literal was spec-correct in isolation. But the gateway's role as a transparent proxy is to handle real client behavior, and the conversion pipeline already has full plumbing to do this for the OpenAI input path. This change makes the two input paths symmetric: the same `normalize_message_roles` (added in #64) and `ensure_alternating_roles` already process the unified-format messages downstream regardless of input shape.

Same model/prompt that 422'd before now flows through cleanly. `claude-opus-4-7` traffic and other paths remain unchanged.

## Test plan
- [x] `pytest tests/unit/test_models_anthropic.py::TestAnthropicMessageRole` — 4/4 pass (user/assistant/system accepted, unknown roles like `developer` still fail fast)
- [x] `pytest tests/unit/test_converters_anthropic.py::TestAnthropicToKiro::test_inline_system_role_is_normalized_to_user` — passes; verifies the resulting Kiro payload's history contains only `userInputMessage`/`assistantResponseMessage` entries
- [x] Regression: `pytest tests/unit/test_models_anthropic.py tests/unit/test_converters_anthropic.py tests/unit/test_converters_core.py tests/unit/test_routes_anthropic.py` — 532/532 pass
- [x] Live verification: 19+ `claude-opus-4-8` requests in journal post-change, all 200; mixed with `4-7` traffic, both work


severity1:
Closing - this branch was based on severity1/main which contained 14 unrelated sync commits, polluting the diff. Reopening from a clean branch off upstream/main with only the fix commit.

PR #197, opened 2026-06-02
fix(anthropic): accept inline system role and merge into system prompt

### LINKED ISSUE #190: BUG: 422 error when newer Claude Code client inlines `role: "system"` in messages array

W0n9:
Fixes #190

Follow-up to the approach discussed in https://github.com/jwadow/kiro-gateway/issues/190#issuecomment-4601707637

## Summary

- Widens `AnthropicMessage.role` to accept `"system"` so Claude Code requests no longer fail Pydantic validation with HTTP 422.
- In `anthropic_to_kiro()`, peels off inline `messages[].role == "system"` entries and merges their text into the top-level `system` prompt (top-level `system` first, inline parts after), then converts only `user` / `assistant` messages.

## Why merge into `system` (not only `normalize_message_roles` → `user`)

For the common Claude Code shape:

```json
[
  {"role": "user", "content": "..."},
  {"role": "system", "content": "<SessionStart / hook context>"}
]
```

normalizing inline `system` to `user` can turn hook context into the current user turn and trigger `ensure_alternating_roles()` placeholders. Merging into `system_prompt` matches the OpenAI path (`role: system` → `system_prompt`, not history) and keeps system instructions separate from user turns.

## Trade-off: prompt caching

Merging changes the serialized prefix vs. client-placed `cache_control` breakpoints, so Anthropic-style prompt cache hits may become misses (cost/latency, not a functional error). The gateway already extracts text only and does not forward `cache_control` to Kiro today. The normalize→`user` approach in #195 also reshapes the request and is not cache-neutral.

## Test plan

- [x] `pytest tests/unit/test_routes_anthropic.py::TestMessagesValidation::test_accepts_system_role_in_messages`
- [x] `pytest tests/unit/test_converters_anthropic.py::TestAnthropicToKiro::test_merges_inline_system_message_into_system_prompt`
- [x] `pytest tests/unit/test_converters_anthropic.py::TestAnthropicToKiro::test_merges_top_level_and_inline_system_prompts_in_order`
- [x] `pytest tests/unit/test_routes_anthropic.py tests/unit/test_converters_anthropic.py` — 158 passed

Made with [Cursor](https://cursor.com)

### DIFF
--- a/kiro/converters_anthropic.py
+++ b/kiro/converters_anthropic.py
@@ -435,7 +435,7 @@ def anthropic_to_kiro(
     This is the main entry point for Anthropic → Kiro conversion.
 
     Key differences from OpenAI:
-    - System prompt is a separate field (not in messages)
+    - System prompt is a separate field; inline system messages are merged into it
     - Content can be string or list of content blocks
     - Tool format uses input_schema instead of parameters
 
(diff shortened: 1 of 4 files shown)


W0n9:

W0n9:
@cla-bot check

W0n9:
> Thanks for the PR! 🎉
> 
> 
> 
> Please reply once with:
> 
> ```
> ```
> 

pyro-x:
Confirming this fix works. I hit the same 422 (`Input should be 'user' or 'assistant'`) with Claude Code v2.1.198 + `claude-opus-4-8` — newer Claude Code versions send `role: "system"` inside `messages` whenever they recognize a modern model ID. I applied the same change locally (accept `system` role + merge into the system prompt, mirroring the existing OpenAI converter) and Claude Code works end-to-end, with all existing tests passing. This also resolves #190, #211, #219 and #226 — would be great to see it merged, as every current Claude Code user hits this.

W0n9:
> Thanks for the PR! 🎉
> 
> 
> 
> Please reply once with:
> 
> ```
> ```
> 

W0n9:
@cla-bot check

PR #198, opened 2026-06-02
fix: aliases not resolved when building Kiro API payload

brandflugan:
### Problem

MODEL_ALIASES were applied during model listing and account routing (ModelResolver.resolve()), but `get_model_id_for_kiro()` — used by `build_kiro_payload()` to determine the actual model ID sent to the Kiro API — did not resolve aliases. This caused aliased model names (e.g. `auto-kiro`) to be sent verbatim to the API, resulting in 400 errors: "Invalid model ID or insufficient subscription level."

### Fix

Added an optional `aliases` parameter to `get_model_id_for_kiro()` that resolves aliases before normalization. Updated `converters_openai.py` to pass `MODEL_ALIASES` to the function.

### Changes

- `kiro/model_resolver.py` — `get_model_id_for_kiro()` now accepts `aliases` dict and resolves before normalizing
- `kiro/converters_openai.py` — passes `MODEL_ALIASES` to `get_model_id_for_kiro()`
- `tests/unit/test_model_resolver.py` — added 3 tests covering alias resolution in payload building

### Tested

- `claude-opus-4.6-kiro` alias correctly resolves to `claude-opus-4.6` in API requests
- Backward compatible — existing behavior unchanged when no aliases provided


brandflugan:

PR #199, opened 2026-06-03
fix(models): accept non-standard message roles in Anthropic requests

day0n:
## Problem

Some clients inline a `system` (or other non-standard) role into the `messages` array of an Anthropic `/v1/messages` request, instead of only using the top-level `system` field. A common case is tooling that spawns sub-agents and reuses the messages array to carry a system reminder.

The gateway rejected every such request with a `422` **before** it reached the converter, because the Pydantic schema pinned the role to a strict literal:

```python
```

Observed error:

```
Validation error (422): [{'type': 'literal_error',
  'loc': ['body', 'messages', 1, 'role'],
  'msg': "Input should be 'user' or 'assistant'",
  'input': 'system', ...}]
```

This is surprising because the gateway already knows how to handle these roles further down the pipeline — the request just never got there.

## Fix

Relax `AnthropicMessage.role` from `Literal["user", "assistant"]` to a plain `str`, matching the already-permissive OpenAI-side schema (`role: str`).

No converter changes are needed. The downstream pipeline in `converters_core.py` already normalizes non-standard roles:

- `normalize_message_roles()` folds any unknown role (e.g. `system`, `developer`) into `user`.
- `ensure_alternating_roles()` then inserts synthetic assistant turns so the Kiro API's alternating user/assistant requirement is preserved.

So the only thing standing in the way was the input schema rejecting the request too early.

## Tests

- Added `TestAnthropicMessageRole` in `test_models_anthropic.py` covering `user`, `assistant`, `system`, `developer`, and a full request with an inlined `system` message.
- Added `test_converts_inlined_system_message` in `test_converters_anthropic.py` verifying the `system` message is converted to a `UnifiedMessage` with role/text preserved (normalization to `user` happens later in the pipeline).
- Updated the route-level test (`test_validates_invalid_role` → `test_accepts_nonstandard_role`) to assert that a non-standard role is **not** rejected with `422`.

Full suite passes: **1697 passed**.

### DIFF
--- a/kiro/models_anthropic.py
+++ b/kiro/models_anthropic.py
@@ -183,11 +183,15 @@ class AnthropicMessage(BaseModel):
     Message in Anthropic format.
 
     Attributes:
-        role: Message role (user or assistant)
+        role: Message role. Officially Anthropic only uses 'user'/'assistant'
+            here, but clients (e.g. Claude Code spawning sub-agents) may inline
+            a 'system' (or other) role into the messages array. We accept any
+            string and let normalize_message_roles() downstream fold unknown
+            roles into 'user', matching the permissive OpenAI-side schema.
         content: Message content (string or list of content blocks)
     """
 
-    role: Literal["user", "assistant"]
+    role: str
     content: Union[str, List[ContentBlock]]
 
     model_config = {"extra": "allow"}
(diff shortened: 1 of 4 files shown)


day0n:

PR #206, opened 2026-06-04
fix(models): accept unknown message roles for Claude Code compatibility

thanhtaivtt:
## Summary
- Relaxed `AnthropicMessage.role` validation from `Literal["user", "assistant"]` to `str` to accept unknown roles (e.g., `system`, `developer`)
- The existing `normalize_message_roles()` converter handles normalization to `user` before forwarding to Kiro API
- Updated test to reflect the new behavior and use realistic `system` role scenario

## Context
Claude Code sends `system` role inside the messages array when proxying through the gateway. Previously this was rejected with 422 at the Pydantic validation layer, before the converter had a chance to normalize it.

## Test plan
- [x] `test_normalizes_unknown_role` passes — unknown roles no longer return 422
- [x] Manual test: proxy Claude Code traffic through gateway, verify `system` role messages go through

### DIFF
--- a/kiro/models_anthropic.py
+++ b/kiro/models_anthropic.py
@@ -187,7 +187,7 @@ class AnthropicMessage(BaseModel):
         content: Message content (string or list of content blocks)
     """
 
-    role: Literal["user", "assistant"]
+    role: str  # Accept any role; converter normalizes unknown roles (e.g., 'system') to 'user'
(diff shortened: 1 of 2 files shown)


thanhtaivtt:
> Thanks for the PR! 🎉
> 
> 
> 
> Please reply once with:
> 
> ```
> ```
> 

PR #175, opened 2026-05-18
fix: handle profileArn runtime endpoint regressions (#168, #173)

el-pablos:
## Summary

Fixes the runtime.kiro.dev/profileArn regression affecting Builder ID/free-tier users and MCP web_search requests.

## Changes

- Include `profileArn` in MCP web_search request bodies when available (closes #173)
- Add `KIRO_USE_LEGACY_ENDPOINT` compatibility fallback for Builder ID accounts (#168)
- Add 10 regression tests for MCP profileArn handling and endpoint config behavior
- Document legacy endpoint fallback and troubleshooting guidance in README and .env.example

## Issues

Fixes #173.
Updates #168 with a documented compatibility workaround.

## Root Cause

Since May 15 2026, `runtime.kiro.dev` requires `profileArn` for ALL account types:
1. MCP `/mcp` endpoint never included `profileArn` in request body → 400 error
2. Builder ID (free tier) accounts have no `profileArn` and no API to discover one → all requests fail

## Verification

```bash
./venv/Scripts/python.exe -m pytest tests/unit/ -v
# Result: 1683 passed, 0 failed, 3 warnings (pre-existing)
```

## Notes

- The legacy endpoint fallback (`KIRO_USE_LEGACY_ENDPOINT=true`) is a compatibility bridge for Builder ID users, not the preferred default.
- MCP endpoint URL is derived from `auth_manager.q_host` which respects the legacy flag automatically.
- No secrets are logged in any affected path.

Issue #173, opened 2026-05-17
BUG: web_search 400


fatim4h:
### Kiro Gateway Version

latest

### What happened?

Try use web search and error, i just vibe coded and add this code after line 138
https://github.com/jwadow/kiro-gateway/blob/b370ec5de14d3cc48cd575f9b71cab1ff9e3832f/kiro/mcp_tools.py#L138

```
 # profileArn is required by runtime.kiro.dev/mcp endpoint
    if auth_manager.profile_arn:
        mcp_request["profileArn"] = auth_manager.profile_arn
```

And work

```
2026-05-18 02:42:13 | INFO     | kiro.streaming_anthropic:stream_kiro_to_anthropic:358 - Intercepted web_search tool call (Path B - MCP emulation)
2026-05-18 02:42:13 | DEBUG    | kiro.streaming_anthropic:stream_kiro_to_anthropic:373 - WebSearch query (Path B): latest news
2026-05-18 02:42:13 | DEBUG    | kiro.mcp_tools:call_kiro_mcp_api:162 - Calling MCP API: https://runtime.us-east-1.kiro.dev/mcp
2026-05-18 02:42:15 | DEBUG    | kiro.mcp_tools:call_kiro_mcp_api:192 - MCP API returned 10 results
2026-05-18 02:42:15 | DEBUG    | kiro.streaming_anthropic:stream_kiro_to_anthropic:689 - [Anthropic Streaming] Completed: input_tokens=2490, output_tokens=16, tool_blocks=0, stop_reason=end_turn
```

### Debug Logs

```
2026-05-18 02:37:27 | INFO     | kiro.streaming_anthropic:stream_kiro_to_anthropic:358 - Intercepted web_search tool call (Path B - MCP emulation)
2026-05-18 02:37:27 | DEBUG    | kiro.streaming_anthropic:stream_kiro_to_anthropic:373 - WebSearch query (Path B): latest news
2026-05-18 02:37:27 | DEBUG    | kiro.mcp_tools:call_kiro_mcp_api:158 - Calling MCP API: https://runtime.us-east-1.kiro.dev/mcp
2026-05-18 02:37:27 | ERROR    | kiro.mcp_tools:call_kiro_mcp_api:164 - MCP API error: 400
2026-05-18 02:37:27 | ERROR    | kiro.streaming_anthropic:stream_kiro_to_anthropic:379 - MCP API call failed for web_search
[...]

PR #213, opened 2026-06-11
fix(mcp): send Kiro client-identity headers on web_search MCP call

severity1:
## Problem

Claude Code's native `web_search` server-side tool, routed through the gateway to `runtime.kiro.dev/mcp`, fails with **403** for IAM Identity Center (AWS SSO OIDC / kiro-cli) accounts, even when MCP is enabled on the profile:

```json
{"message":"User is not authorized to make this call.","reason":null}
```

This 403 is distinct from the `profileArn` **400** issue (#173, addressed by #180/#189/#175). It surfaces only once `profileArn` is present: the request is then authenticated, but the endpoint still declines it.

## Root Cause

`call_kiro_mcp_api()` in `mcp_tools.py` sent only three headers (`Authorization`, `x-amzn-codewhisperer-optout`, `Content-Type`). The completion path (`/generateAssistantResponse`, via `get_kiro_headers` in `utils.py`) sends a full Kiro client-identity set: `User-Agent` with the `KiroIDE-<version>-<fingerprint>` signature, `x-amz-user-agent`, `x-amzn-kiro-agent-mode`, and the `amz-sdk-*` pair.

The `/mcp` endpoint gates authorization on these client-identity signals. A bare `Authorization` header is rejected with 403 even when the account is entitled. Completions work because they send the full identity; MCP did not.

## Fix

Build the MCP headers from the canonical `get_kiro_headers` (single source of truth, shared with the completion path) and override only the three fields that differ for `/mcp`:

- `Content-Type: application/json` — `/mcp` is JSON-RPC, not the `x-amz-json-1.0` completion API
- drop `x-amz-target` — completion-operation-specific, meaningless to `/mcp`
- `x-amzn-codewhisperer-optout: false` — preserves the original MCP value

Reusing `get_kiro_headers` (rather than duplicating the literals) means future `KiroIDE` version bumps flow to the MCP call automatically instead of silently drifting.

## Scope — depends on profileArn contributions

This PR is deliberately surgical: it contains **only** the novel client-identity-headers fix.

A complete `web_search` fix **also requires `profileArn` in the MCP request body** — without it the endpoint returns **400**, before the 403 this PR addresses. That `profileArn` work is **not** included here; it is contributed separately by **#180 / #189 / #175** (and tracked in #173). This change stacks on top of those: once a `profileArn` PR merges, this header fix carries the call the rest of the way past the 403.

Credit to those contributors for the `profileArn` half of the puzzle.

## Testing

- Verified end-to-end: the native `web_search` tool returns live results through the gateway on an enterprise SSO OIDC (kiro-cli) account, where it previously 403'd (with `profileArn` applied locally).
- Added a regression test asserting the MCP request sends the Kiro client-identity headers.
- `tests/unit/test_mcp_tools.py` passes (22 tests).

The header change is additive (sends more identity signal, matching the completion path), so it should not regress auth types that already worked.

### DIFF
--- a/kiro/mcp_tools.py
+++ b/kiro/mcp_tools.py
@@ -41,6 +41,7 @@ from fastapi.responses import JSONResponse, StreamingResponse
 from loguru import logger
 
 from kiro.tokenizer import count_message_tokens, count_tokens
+from kiro.utils import get_kiro_headers
 
 # Import debug_logger
 try:
@@ -135,7 +136,7 @@ async def call_kiro_mcp_api(
             "arguments": {"query": query}
         }
     }
-    
+
     # Log MCP request
     try:
         mcp_request_json = json.dumps(mcp_request, ensure_ascii=False, indent=2).encode('utf-8')
@@ -146,13 +147,13 @@ async def call_kiro_mcp_api(
     
     try:
         token = await auth_manager.get_access_token()
-        
-        # EXACT headers from architecture
-        headers = {
-            "Authorization": f"Bearer {token}",
-            "x-amzn-codewhisperer-optout": "false",
-            "Content-Type": "application/json"
-        }
+
+        # /mcp requires the same Kiro client-identity headers as the completion
+        # path (else 403); reuse get_kiro_headers and override what differs.
+        headers = get_kiro_headers(auth_manager, token)
+        headers["Content-Type"] = "application/json"  # /mcp is JSON-RPC
+        headers.pop("x-amz-target", None)
+        headers["x-amzn-codewhisperer-optout"] = "false"
         
         mcp_url = f"{auth_manager.q_host}/mcp"
         logger.debug(f"Calling MCP API: {mcp_url}")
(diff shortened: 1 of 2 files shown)


LazarusX:
Hey, I ran into the same issue and combined this PR's header fix with the profileArn fix from #175/#180. One thing I found that neither PR addresses: if `auth_manager.profile_arn` is empty at call time (e.g. token loaded from cache, no refresh triggered yet), you need to force a refresh to get it from the response. Something like:

```python
profile_arn = auth_manager.profile_arn or PROFILE_ARN or ""
if not profile_arn:
    token = await auth_manager.force_refresh()
    profile_arn = auth_manager.profile_arn or PROFILE_ARN or ""
```

Without this, accounts that haven't done a token refresh since startup will still hit the 400 even with profileArn code in place. Might be worth adding to this PR or #175 since they stack together anyway.

severity1:
@LazarusX i think it makes sense to stack it against #175 since it deals with the same parts of the code.

PR #215, opened 2026-06-11
fix(converters): strip Claude Code 2.1.x billing attribution from sys…

liwenquantop-dot:
## Problem

Claude Code 2.1.x prepends a per-request system text block of the form:

x-anthropic-billing-header: cc_version=2.1.153.d02; cc_entrypoint=sdk-cli; cch=<5hex>;

The `cch` token is a fresh random hex on every request. Because `extract_system_prompt()` concatenates all system text blocks into a single string before forwarding to Kiro, this random prefix poisons the prompt seen by upstream and invalidates any prefix-keyed prompt cache. (The attribution header is meant to be an HTTP header but ships in the request body for `claude -p` / SDK mode.)

## Solution

Add `_strip_billing_attribution()` and call it from `extract_system_prompt()` for both string and list system formats. Pure-attribution blocks are dropped; mixed blocks have the leading attribution line removed. Behavior is gated by `STRIP_BILLING_HEADER` (default `true`) so it can be disabled for A/B comparison.

## Empirical validation (local)

Replay benchmark using real captured Claude Code body (212KB, 3 system blocks), N=30 interleaved, randomized cch per request:

| Metric | Baseline (no strip) | Patched (strip) | Δ |
| --- | --- | --- | --- |
| TTFB mean | 4813ms | 4238ms | **-574ms (-11.9%)** |
| TTFB median | 4506ms | 3952ms | -554ms (-12.3%) |
| TTFB p95 | 6263ms | 4958ms | -1305ms (-20.8%) |

Byte-level check via `DEBUG_MODE=all`: incoming body contains `cch=<hex>` and `x-anthropic-billing-header`; outgoing Kiro payload has zero matches for either string.

## Alternative

`CLAUDE_CODE_ATTRIBUTION_HEADER=0` set in the client environment achieves the same effect on the Claude Code side. The gateway-level fix is complementary and protects users / clients that don't (or can't) set the env var.

## Files

- `kiro/config.py`: add `STRIP_BILLING_HEADER` env (default true)
- `kiro/converters_anthropic.py`: add `_strip_billing_attribution()`, apply in `extract_system_prompt()`

### DIFF
--- a/kiro/config.py
+++ b/kiro/config.py
@@ -343,6 +343,19 @@ TRUNCATION_RECOVERY: bool = os.getenv("TRUNCATION_RECOVERY", "true").lower() in 
 LOG_LEVEL: str = os.getenv("LOG_LEVEL", "INFO").upper()
 
 # ==================================================================================================
+# Billing Attribution Strip
+# ==================================================================================================
+
+# Claude Code 2.1.x prepends a per-request system text block of the form
+# ``x-anthropic-billing-header: cc_version=...; cc_entrypoint=...; cch=<5hex>;``
+# where ``cch`` is a fresh random hex token. The Kiro gateway concatenates all
+# system text blocks before forwarding, so this random prefix would invalidate
+# any upstream prompt cache keyed on the prompt prefix. When enabled (default),
+# the gateway strips the leading attribution line / block before forwarding.
+# Disable only for A/B comparison or to debug attribution behavior.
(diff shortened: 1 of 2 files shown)


liwenquantop-dot:

liwenquantop-dot:
@jwadow  Please add `liwenquantop-dot` to the `.clabot` contributors list.
I have read the CLA and accept its terms.

PR #216, opened 2026-06-12
Strip leaked native tool-call markup from assistant content

docooler:
Some upstream models (notably deepseek) emit their native tool-call serialization (e.g. `<｜DSML｜function_calls...`) as raw text in the assistant `content`, even though the gateway already extracts the structured `tool_calls` separately. The duplicated markup then leaks into the OpenAI-format response `content`.

Reproduced deterministically against deepseek-3.2: whenever the model emits a tool call, `content` contains the DSML markup regardless of whether `tools` were declared in the request (structured `tool_calls` are parsed correctly in both cases).

Fix: add `strip_tool_markup()` and apply it to the accumulated content in `collect_stream_response()` before building the message, cutting from the first known marker to end-of-string. `tool_calls` are untouched.

Note: covers the non-streaming path; the streaming path needs cross-chunk buffering to strip markers split across deltas (follow-up).

### DIFF
--- a/kiro/streaming_openai.py
+++ b/kiro/streaming_openai.py
@@ -69,6 +69,34 @@ except ImportError:
 __all__ = ['FirstTokenTimeoutError', 'stream_kiro_to_openai', 'stream_with_first_token_retry', 'collect_stream_response']
 
 
+# Native tool-call serialization markers that some models (notably deepseek)
+# leak into the assistant text content. Structured tool_calls are parsed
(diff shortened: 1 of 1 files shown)


docooler:

docooler:
@jwadow I've signed the CLA above. Could you please add docooler to the .clabot contributors list so the check can pass? Thanks!

docooler:

PR #218, opened 2026-06-13
docs: add Windows kiro-cli env example

chaman2003:
Adds the default Windows kiro-cli SQLite database path to .env.example.

This helps Windows users copy .env.example to .env and configure KIRO_CLI_DB_FILE correctly without accidentally using the Linux/macOS path. It can resolve cases where the gateway keeps reading stale credentials or an old Kiro account after logging in with kiro-cli on Windows.


chaman2003:

PR #229, opened 2026-06-21
docs: add Windows kiro-cli SQLite DB path to .env.example

chaman2003:
## Summary

- Adds the default Windows `kiro-cli` SQLite database path to `.env.example` so Windows users can copy it directly without guessing the path
- Renames the legacy `KIRO_CREDS_FILE` env var to `KIRO_CLI_DB_FILE` for consistency across all docs, config, and code
- Updates all language-specific READMEs, architecture docs, docker-compose, and tests to reflect the rename

## Motivation

Windows users copying `.env.example` to `.env` were likely to use the Linux/macOS path or the old `KIRO_CREDS_FILE` variable name, which caused the gateway to read stale credentials or an old Kiro account after logging in with `kiro-cli` on Windows.

## Test plan

- [ ] Verify `KIRO_CLI_DB_FILE` is accepted in `.env` on Windows and the gateway picks up the correct SQLite database
- [ ] Confirm no references to `KIRO_CREDS_FILE` remain in docs or runnable code
- [ ] Run existing test suite (`pytest`) to confirm no regressions

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,7 +31,10 @@ PROXY_API_KEY="my-super-secret-password-123"
 
 # Path to kiro-cli SQLite database (for AWS IAM Identity Center users)
 # The gateway will auto-detect AWS SSO OIDC and use the correct endpoint
+# Linux/macOS:
 # KIRO_CLI_DB_FILE="~/.local/share/kiro-cli/data.sqlite3"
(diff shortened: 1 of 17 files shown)


chaman2003:

PR #232, opened 2026-06-23
fix(converters): resolve top-level oneOf/allOf/anyOf in tool schemas

xy69:
## Summary

Fixes Bedrock `TOOL_SCHEMA_INVALID` error when tools have `oneOf`, `allOf`, or `anyOf` at the top level of their `input_schema`.

## Problem

OpenCode (and potentially other clients) send tools whose input schemas use JSON Schema composition keywords at the root level. Bedrock rejects these with:
```
tools.67.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level (reason: TOOL_SCHEMA_INVALID)
```

## Solution

Added `_resolve_top_level_composition()` that flattens these before sending to Kiro:
- **anyOf/oneOf**: picks the first object-type variant (or first if none is object)
- **allOf**: merges all variants into a single schema

Nested composition keywords are preserved — Bedrock only rejects them at the root.

## Testing

- 8 new tests covering all resolution strategies and edge cases
- All 17 `TestSanitizeJsonSchema` tests pass
- Full suite: 1679 passed, no regressions

### DIFF
--- a/kiro/converters_core.py
+++ b/kiro/converters_core.py
@@ -436,24 +436,120 @@ def inject_thinking_tags(content: str, thinking_config: ThinkingConfig) -> str:
 # JSON Schema Sanitization
 # ==================================================================================================
 
-def sanitize_json_schema(schema: Optional[Dict[str, Any]]) -> Dict[str, Any]:
+def _resolve_top_level_composition(schema: Dict[str, Any]) -> Dict[str, Any]:
+    """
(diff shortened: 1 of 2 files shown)


xy69:

ankitcharolia:
@xy69 could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

Issue #226, opened 2026-06-18
BUG: Claude Code 2.1.168 sends system role causing 422 validation error


HARSHP55555:
### Kiro Gateway Version

v2.4.dev.13

### What happened?

Environment
Kiro Gateway: v2.4.dev.13
Claude Code: 2.1.168
OS: Windows 11
Problem

After upgrading Claude Code, all requests fail with a 422 validation error.

Kiro Gateway starts successfully and authenticates correctly, but requests to /v1/messages?beta=true are rejected.

Error:

Validation error (422):
Input should be 'user' or 'assistant'
input: 'system'

Server log:

POST /v1/messages?beta=true HTTP/1.1" 422

Validation error (422):
[
  {
    "type": "literal_error",
    "loc": ["body", "messages", 1, "role"],
    "msg": "Input should be 'user' or 'assistant'",
    "input": "system"
  }
]
Investigation

The request payload contains a message with:

{
  "role": "system",
  "content": "..."
}

However the gateway model currently validates:

class AnthropicMessage(BaseModel):
    role: Literal["user", "assistant"]

which causes validation to fail before the request is processed.

Suggested Fix

Allow system role messages, e.g.:

role: Literal["user", "assistant", "system"]

or consider using:

role: str

if newer Claude Code versions may introduce additional roles.

Reproduction
Start Kiro Gateway v2.4.dev.13
Configure Claude Code to use the gateway
Run claude
Enter a simple prompt such as hi
Observe 422 validation error

### Debug Logs

● API Error: 422 {"detail":[{"type":"literal_error","loc":["body","messages",1,"role"],"msg":"Input should be 'user' or 'assistant'","input":"system","ctx":{"expected":"'user' or 'assistant'"}}],"body":"{\"model\":\"claude-opus-4-8\",\"messages\":[{\"role\":\"user\",\"content\":[{\"type\":\"text\",\"text\":\"<system-reminder>\\nAs you answer the user's questions, you can use the following context:\\n# currentDate\\nToday's date is 2026-06-18.\\n\\n      IMPORTANT: this context may or may not be relevant to your tasks. You should not respond to this context unless it is highly relevant to your task.\\n</system-reminder>\\n\\n\"},{\"type\":\"text\",\"text\":\"hi\",\"cache_control\":{\"type\":\"ephemeral\"}}]},{\"role\":\"system\",\"content\":\""}
-----------------------------------------------------------------------------------------------------------------------------------

[...]

Issue #219, opened 2026-06-15
BUG: Validation error and extraction for Anthropic system messages


balraam1:
### Kiro Gateway Version

latest

### What happened?

When using the Anthropic API and providing messages with the "system" role in the `messages` array, the gateway raises a 422 Validation Error (`Input should be 'user' or 'assistant'`) and fails. 

The gateway should allow system messages in the `messages` array for Anthropic, extract them, and merge them with the main `system_prompt` properly without triggering validation errors.

### Debug Logs

2026-06-04 11:36:20 | ERROR    | kiro.exceptions:validation_exception_handler:88 - Validation error (422): [{'type': 'literal_error', 'loc': ['body', 'messages', 1, 'role'], 'msg': "Input should be 'user' or 'assistant'", 'input': 'system', 'ctx': {'expected': "'user' or 'assistant'"}}]

(Note: No files from debug_logs/ are attached because the request failed at Pydantic validation before reaching Kiro, so no debug logs were generated.)

ankitcharolia:
@balraam1 would you like to give a try to this kiro-gateway:

https://github.com/ankitcharolia/kiro-gateway (it is using ACP compliant way to talk to kiro-cli binary). It works fine with multiple AI harnesses. I also faced the same issue, so I refactored it completely.

Issue #211, opened 2026-06-09
BUG: claude code使用用 claude-opus-4-8 报错 422


zqiyu:
### Kiro Gateway Version

Version: 0.12.301

### What happened?

Pydantic 放行 system role，进入业务逻辑后把这些消息的文本剥离，合并到顶层 system
  字段。

### Debug Logs

[claude-code原始请求.json](https://github.com/user-attachments/files/28746336/claude-code.json)

colinwang-wang:
现在kiro-gateway 本地认证失败，解决了么？

ankitcharolia:
@zqiyu @colinwang-wang  would you like to give a try to this kiro-gateway:

https://github.com/ankitcharolia/kiro-gateway (it is using ACP compliant way to talk to kiro-cli binary). It works fine with multiple AI harnesses. I also faced the same issue, so I refactored it completely.

PR #235, opened 2026-06-26
Fix Claude Code / Opus 4.8 compatibility and web_search reliability

### LINKED ISSUE #190: BUG: 422 error when newer Claude Code client inlines `role: "system"` in messages array

### LINKED ISSUE #208: BUG: models not include claude-opus-4.8
### Kiro Gateway Version

v2.3

### What happened?

/models接口里没有4.8模型

### Debug Logs

Image

### LINKED ISSUE #231: BUG: web search issue
### Kiro Gateway Version

latest

### What happened?

web search isnt working properly

### Debug Logs

web search aint working

coderhisham:
This PR bundles several fixes that make the gateway work reliably with the latest Claude Code (Opus 4.8), plus quality-of-life improvements for running the server. All changes keep the gateway's pass-through/transparent philosophy, are consistent across both the OpenAI and Anthropic surfaces (and streaming + non-streaming), and ship with tests. Full suite passes (only 2 environment-specific `.env` default-host/port tests fail locally, unrelated to these changes).

## What's included

**1. `feat(models)`: Claude Opus 4.8 (1M context) + configurable context windows — closes #208**
- Adds `claude-opus-4.8` to `FALLBACK_MODELS` with a 1M `maxInputTokens`, so it's recognized on the `runtime.kiro.dev` endpoint (which has no `/ListAvailableModels`).
- Makes context accounting configurable: `DEFAULT_MAX_INPUT_TOKENS` is now an env var, and a new `MODEL_CONTEXT_WINDOWS` JSON env lets users declare per-model windows that take precedence over discovered/default limits. Affects token-usage estimation only; requests stay pass-through.

**2. `fix(anthropic)`: accept inline `system` role + server-side web_search blocks — closes #190, #219 (also #211, #226)**
- Claude Code injects inline `role: "system"` messages, which the strict `Literal["user","assistant"]` rejected with 422. `AnthropicMessage.role` is now a free-form string (mirroring the OpenAI model); inline system messages are hoisted into the top-level system prompt, matching the OpenAI adapter and avoiding spurious user turns.
- After a web_search, follow-up turns echo back `server_tool_use` / `web_search_tool_result` blocks. These are now part of the `ContentBlock` union, and their content is folded into text so search grounding survives conversion.

**3. `fix(mcp)`: reliable web_search, never leak the tool call — closes #231, #173**
- The `/mcp` call now sends the full Kiro client-identity headers (it previously sent only `Authorization`, causing 403) and includes top-level `profileArn` for Enterprise/Builder accounts.
- On any MCP failure, the gateway no longer leaks a raw `web_search` tool_use/tool_call to the client (which surfaced as "No such tool available: web_search"). It now degrades gracefully — Anthropic emits a native `web_search_tool_result` error; OpenAI emits an in-band "unavailable" note — across streaming and non-streaming.

**4. `fix(auth)`: actionable 401 on invalid/expired refresh token**
- A rejected refresh token (e.g. AWS SSO OIDC `invalid_grant`) previously surfaced as a generic 500. It now raises a typed error mapped to a clear 401 with re-auth guidance in both routes. Transient/server errors still propagate unchanged.

**5. `feat(cli)`: `--stop` and `--force` to free the server port**
- `--stop` stops any process listening on the resolved port and exits; `--force` stops the occupant and starts anyway. Adds a preflight check so an occupied port gives an actionable message instead of a raw "address already in use" traceback. Cross-platform (`lsof`/`netstat`), only targets processes listening on the exact port, never the current process.

**6. `fix(server)`: bounded graceful shutdown so Ctrl+C frees the port promptly**
- With long-lived streaming, uvicorn's graceful shutdown could hang on open SSE connections and keep the port bound. Adds `SHUTDOWN_TIMEOUT` (env, default 10s) passed as `timeout_graceful_shutdown`. A second Ctrl+C still forces immediate exit.

## Testing
- New/updated unit tests for every change: models validation, converters (system-role hoisting + echoed web_search blocks), MCP headers/profileArn, web_search anti-leak across both APIs and both modes, auth 401 detection, port utilities, CLI flags, and config.
- Network-isolated per the project's test philosophy.

## Notes
- New env vars documented in `.env.example`: `DEFAULT_MAX_INPUT_TOKENS`, `MODEL_CONTEXT_WINDOWS`, `SHUTDOWN_TIMEOUT`.
- Happy to split this into per-concern PRs if you'd prefer to review/merge them independently.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -213,6 +213,12 @@ PROXY_API_KEY="my-super-secret-password-123"
 # Default: 300 seconds (5 minutes) - generous timeout to avoid premature disconnects.
 # STREAMING_READ_TIMEOUT="300"
 
+# Graceful shutdown timeout (in seconds) for Ctrl+C / SIGTERM.
+# Caps how long the server waits for in-flight (streaming) connections to drain
+# before exiting, so Ctrl+C frees the port promptly instead of hanging on an open
+# SSE stream. A second Ctrl+C forces an immediate exit. Default: 10 seconds.
+# SHUTDOWN_TIMEOUT="10"
+
 # ===========================================
 # FAKE REASONING (Extended Thinking via Tag Injection)
 # ===========================================
--- a/.env.example
+++ b/.env.example
@@ -172,6 +172,25 @@ PROXY_API_KEY="my-super-secret-password-123"
 # STATE_SAVE_INTERVAL_SECONDS=10
 
 # ===========================================
+# CONTEXT WINDOW (Token Accounting)
+# ===========================================
+
+# Default max input tokens used for token-usage estimation when the upstream
+# model limit is unknown. This is NOT an enforced cap - requests are always
+# passed through to Kiro, which decides the real limit. It only affects the
+# input_tokens reported in usage (context_usage_percentage * this value).
+# Default: 200000
+# DEFAULT_MAX_INPUT_TOKENS=200000
+
+# Per-model context window overrides (JSON object: model id -> max input tokens).
+# Highest precedence - wins over both the discovered /ListAvailableModels value
+# and DEFAULT_MAX_INPUT_TOKENS. Use this to declare a model's real window (e.g.
+# 1M on Kiro) when the gateway cannot discover it - notably on the
+# runtime.kiro.dev endpoint, which does not provide /ListAvailableModels.
+# Keys are Kiro model ids (after normalization), e.g. "claude-sonnet-4".
+# MODEL_CONTEXT_WINDOWS='{"claude-sonnet-4": 1000000, "claude-opus-4.7": 1000000}'
+
+# ===========================================
 # FIRST TOKEN TIMEOUT (Streaming Retry)
 # ===========================================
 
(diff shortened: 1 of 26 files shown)


ankitcharolia:
@coderhisham could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #237, opened 2026-06-29
fix(mcp): route web_search to Amazon Q endpoint with profile ARN for AWS SSO OIDC accounts

yeeyon:
## Problem

`web_search` (via the Kiro MCP endpoint) fails for **kiro-cli / AWS SSO OIDC** accounts.

`call_kiro_mcp_api` sends the request to `https://runtime.{region}.kiro.dev/mcp` (the `q_host`) without a profile ARN. For AWS SSO OIDC accounts that endpoint rejects the call:

1. `400 Bad Request` - `profileArn is required`
2. after adding the ARN to the body, `403 Forbidden` - `User is not authorized`

The MCP call then returns `None`, and clients (e.g. Claude Code) hang retrying the web_search tool. Kiro Desktop accounts are unaffected because their auth context resolves differently, which is why this is easy to miss.

## Root cause

For AWS SSO OIDC accounts the MCP `web_search` tool must be called on the **Amazon Q** endpoint, with the profile ARN supplied as a header (not in the body), and with the `x-amz-json-1.0` content type:

| | Before (broken for SSO) | After |
|---|---|---|
| Host | `https://runtime.{region}.kiro.dev/mcp` | `https://q.{region}.amazonaws.com/mcp` |
| profile ARN | not sent | header `x-amzn-kiro-profile-arn` |
| Content-Type | `application/json` | `application/x-amz-json-1.0` |

## Fix

- For `AWS_SSO_OIDC` auth with a profile ARN, route `web_search` to `https://q.{region}.amazonaws.com/mcp`, send `x-amzn-kiro-profile-arn`, and use `application/x-amz-json-1.0`. Region is derived from the profile ARN.
- **Kiro Desktop auth keeps the existing `q_host` behavior** (no change).
- The non-200 error log now includes the response body so future MCP failures are not silent.

## Tests

Adds `TestMCPEndpointSelection` with two cases:
- AWS SSO OIDC -> Amazon Q host + `x-amzn-kiro-profile-arn` header + `x-amz-json-1.0`
- Kiro Desktop -> unchanged `q_host`, no profile-ARN header

`pytest tests/unit/test_mcp_tools.py` -> 23 passed.

## Verification

Verified end-to-end against a live kiro-cli (AWS SSO OIDC) account: native `web_search` now streams `server_tool_use` -> `web_search_tool_result` -> results, where it previously hung.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,9 @@ requests/
 .pytest_cache/
 .coverage
 htmlcov/
+
+# Local gateway runtime artifacts (may contain tokens/ARNs - never commit)
+gateway.pid
+gateway.log
+gateway.out.log
+gateway.err.log
--- a/kiro/mcp_tools.py
+++ b/kiro/mcp_tools.py
@@ -146,22 +146,44 @@ async def call_kiro_mcp_api(
     
     try:
         token = await auth_manager.get_access_token()
-        
-        # EXACT headers from architecture
-        headers = {
-            "Authorization": f"Bearer {token}",
-            "x-amzn-codewhisperer-optout": "false",
-            "Content-Type": "application/json"
-        }
-        
-        mcp_url = f"{auth_manager.q_host}/mcp"
+
+        # Endpoint selection for web_search MCP:
+        # - kiro-cli / AWS SSO OIDC accounts must call the Amazon Q endpoint
+        #   (https://q.{region}.amazonaws.com/mcp) and pass the profile ARN via the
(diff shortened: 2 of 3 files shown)


yeeyon:

yeeyon:
@cla-bot check

ankitcharolia:
@yeeyon could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #238, opened 2026-06-29
fix(converters): use "." for empty content and merge same-role turns

obodnikov:
## Problem
In agentic tool-use loops the model receives fabricated placeholder text it
interprets as genuine input — it remarks on "the user sent an empty message"
or reacts to the literal `(empty placeholder)` string between tool executions.

Two sources in `converters_core.py`:
1. Empty `content` (tool-result-only user turns, tool-call-only assistant turns)
   was filled with `(empty placeholder)` / `[System: ...]`.
2. `ensure_alternating_roles()` inserted synthetic `assistant: "(empty placeholder)"`
   turns between consecutive user messages.

Both land in `content`, which is the model's actual utterance channel — so the
model reads/echoes them.

## Fix
Adopt the LiteLLM convention (PR BerriAI/litellm#28987), adapted for Kiro's
non-empty-`content` requirement:
- `EMPTY_CONTENT_PLACEHOLDER = "."` for all empty-content fallbacks. A bare `.`
  can't be misread as intent or echoed; tool payloads stay in `toolUses`/`toolResults`.
- `ensure_alternating_roles()` **merges** consecutive same-role messages
  (content, tool_calls, tool_results, images) instead of fabricating turns.
- Skip thinking-tag injection for tool-result-only current messages.

`"(empty result)"` (empty *tool output*) is a separate, legitimate case and is
left unchanged.

## Why this approach
LiteLLM's older `[System: ...]` placeholder caused this exact leak/echo
(BerriAI/litellm#24498 — the model even reproduced it with a typo, proving it
was repeating history). PR #28987 replaced it with `.` + "never put diagnostics
in `content`, preserve tool content." Bedrock Converse (#20364) and the Anthropic
API take the same "drop/merge, don't fabricate" stance.

## Testing
`pytest tests/unit/test_converters_core.py tests/unit/test_converters_openai.py`
→ 322 passing (covers both OpenAI and Anthropic adapters; conversion is shared
and identical for streaming/non-streaming).

### DIFF
--- a/kiro/converters_core.py
+++ b/kiro/converters_core.py
@@ -48,6 +48,21 @@ from kiro.payload_guards import check_payload_size, trim_payload_to_limit
 
 
 # ==================================================================================================
+# Constants
+# ==================================================================================================
+
+# Filler used when a message has no text but the Kiro API still requires a non-empty `content`
+# string (the real payload lives in toolUses / toolResults). A bare "." is deliberately minimal:
+# unlike descriptive fillers such as "(empty placeholder)" or "[System: ...]", it cannot be
+# misread by the model as user/assistant intent, and it does not get echoed back across turns.
+#
+# This mirrors LiteLLM's `_DEFAULT_EMPTY_CONTENT_MESSAGE = "."` (BerriAI/litellm PR #28987), which
+# replaced the older "[System: ...]" placeholders precisely because those leaked into `content`
+# and the model started repeating them on subsequent turns.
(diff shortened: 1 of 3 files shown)


obodnikov:

obodnikov:

ankitcharolia:
@obodnikov could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #242, opened 2026-07-04
fix(config): add claude-opus-4.8 to FALLBACK_MODELS

fengseekling-coder:
## What

- Adds `claude-opus-4.8` to `FALLBACK_MODELS` so the fallback list stays
  in sync with the model catalog (4.5/4.6/4.7 are already present).
- Adds `.venv/` and `.DS_Store` to `.gitignore` — these are dev-machine
  artifacts that show up in `git status` and add noise.

## Why

- claude-opus-4.8 should match the same upstream flag pattern as the
  rest of the catalog.
- `.gitignore` was missing the two most common Python-on-macOS
  artifacts.

## Test

- `python -c "from kiro.config import FALLBACK_MODELS; print(any(m['modelId']=='claude-opus-4.8' for m in FALLBACK_MODELS))"`
  → `True`
- `git status` no longer reports `.venv/` or `.DS_Store` as untracked.

Co-authored-by: Cursor <cursoragent@cursor.com>

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,11 @@ state.json
 .vscode/
 .idea/
 .shard/
+.DS_Store
+
+# Python venv
+.venv/
+venv/
 
 # Python
 __pycache__/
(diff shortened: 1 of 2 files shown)


fengseekling-coder:

ankitcharolia:
@fengseekling-coder could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

fengseekling-coder:
Thanks for the recommendation, Ankit. I'll take a look at it.

PR #244, opened 2026-07-07
Enhance Anthropic message conversion and update model mappings

Harsh502s:
Enhance Anthropic message conversion and update model mappings


ankitcharolia:
@Harsh502s could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #246, opened 2026-07-09
fix(models): accept opus-4.8 and hoist inline system messages

sour2001:
## What & Why

Two related fixes so `claude-opus-4.8` works end-to-end through the gateway. Both surfaced as `422` errors when pointing a client (Claude Code) at the Anthropic endpoint.

### 1. Add `claude-opus-4.8` to `FALLBACK_MODELS`
The fallback model list topped out at `claude-opus-4.7`. When Kiro's `/ListAvailableModels` doesn't return 4.8 (or the gateway falls back to the static list), the resolver treated it as an unknown pass-through model, which Kiro then rejected. Added the ID so it resolves from cache like the other current models. The name normalizer already handles the common input shapes (`claude-opus-4-8`, `claude-4.8-opus-high`, date suffixes) — they all map to `claude-opus-4.8`.

### 2. Hoist inline `role: "system"` messages
Anthropic's schema only accepts `user`/`assistant` inside the `messages` array; the system prompt belongs in the top-level `system` field. Some clients (Claude Code, LangChain, OpenAI-style callers) inline `role: "system"` messages instead, which failed with:

```
422 literal_error: Input should be 'user' or 'assistant' ... input: "system"
```

Added a `mode="before"` validator on `AnthropicMessagesRequest` and `AnthropicCountTokensRequest` that moves inline system messages into the top-level `system` field before per-message validation runs. It:
- Merges with an existing top-level `system` (string or list-of-blocks form) rather than clobbering it
- Joins multiple inline system messages in order
- Handles system content given as a string or as a list of text blocks
- Strips empty inline system messages (which would otherwise still 422)
- Is a no-op when there are no inline system messages

Applied to **both** request models for consistency (messages + count_tokens).

## Test coverage

All new tests follow the existing Arrange-Act-Assert + print-diagnostics style.

- `tests/unit/test_models_anthropic.py::TestInlineSystemMessageHoisting` — 10 tests: single/multiple inline system messages, merge with existing string/list system, block-form content, empty content, count_tokens endpoint, and typed-object passthrough.
- `tests/unit/test_config.py::TestFallbackModelsIntegration::test_opus_4_8_present_and_resolves` — verifies `claude-opus-4.8` is in the list and resolves from cache for both dot and dash client formats.

Full suite: **1702 passed**.

### DIFF
--- a/kiro/config.py
+++ b/kiro/config.py
@@ -282,6 +282,7 @@ FALLBACK_MODELS: List[Dict[str, str]] = [
     {"modelId": "claude-opus-4.5"},
     {"modelId": "claude-opus-4.6"},
     {"modelId": "claude-opus-4.7"},
+    {"modelId": "claude-opus-4.8"},
     {"modelId": "deepseek-3.2"},
     {"modelId": "glm-5"},
     {"modelId": "minimax-m2.1"},
--- a/kiro/models_anthropic.py
+++ b/kiro/models_anthropic.py
@@ -339,6 +339,72 @@ class AnthropicMessagesRequest(BaseModel):
 
     model_config = {"extra": "allow"}
 
+    @model_validator(mode="before")
+    @classmethod
+    def _hoist_system(cls, data: Any) -> Any:
+        return _hoist_system_messages(data)
+
+
+def _hoist_system_messages(data: Any) -> Any:
+    """
+    Move any inline ``role: "system"`` messages out of ``messages`` and into
+    the top-level ``system`` field. Some clients (Claude Code, LangChain,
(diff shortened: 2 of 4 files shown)


sour2001:

musa-2066462 (approved):
I just tested this code locally and it works perfectly. 
U just need to change the anthropic model to this : 
export ANTHROPIC_MODEL="claude-opus-4-8-20260528"

ankitcharolia:
@sour2001 @musa-2066462 could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

It supports all necessary endpoints for OpenAI and Anthropic.

PR #210, opened 2026-06-07
fix: accept system role in AnthropicMessage to support Opus Claude Code

coder-bks:
## Problem
Claude Code sends messages with `role: "system"` as part of its 
standard request format. The current Pydantic model restricts role 
to `Literal["user", "assistant"]`, causing a 422 validation error 
on every request.

## Fix
Relaxed the role field from `Literal["user", "assistant"]` to `str` 
to accept all role values including system.

## Tested
- Windows, Claude Code v2.1.167
- Confirmed 422 for Opus error before fix, resolved after

PR #207, opened 2026-06-05
fix(anthropic): allow and extract system messages inside messages list

balraam1:
Some clients/wrappers send system prompts inside the messages array under the system role when targeting /v1/messages. This PR updates the Anthropic schema to accept role: "system", extracts them from the message history list, and merges them directly into the system prompt parameter (matching the existing behavior of the OpenAI converter).

Fixes #219

PR #249, opened 2026-07-17
fix(models): accept non-standard roles in AnthropicMessage so normalization can run

### LINKED ISSUE #190: BUG: 422 error when newer Claude Code client inlines `role: "system"` in messages array

### LINKED ISSUE #226: BUG: Claude Code 2.1.168 sends system role causing 422 validation error

Navapon:
## What

Widens `AnthropicMessage.role` from `Literal["user", "assistant"]` to `str` in `kiro/models_anthropic.py`, so requests containing inline non-standard roles (e.g. `role: "system"` sent by newer Claude Code clients for SessionStart hook context) are accepted and handled by the gateway's existing normalization pipeline instead of being rejected with:

```
422 {"detail":[{"type":"literal_error","loc":["body","messages",N,"role"],"msg":"Input should be 'user' or 'assistant'", ...}]}
```

Fixes #190
Fixes #226

## Why this approach

- **The handling logic already exists** — `normalize_message_roles()` in `converters_core.py` (added for issue #64) coerces any non-user/assistant role to `user` before `ensure_alternating_roles()`. The strict `Literal` on the Anthropic model rejects the request *before* that logic can run. This PR removes the gate rather than adding a new mechanism, handling the entire class of unknown-role issues ("system", "developer", future roles), not just one value.
- **Consistency across both APIs** (per CONTRIBUTING.md): the OpenAI path already does exactly this — `ChatMessage.role` is `str` (`kiro/models_openai.py:78`), and `tests/unit/test_routes_openai.py::test_validates_invalid_role` already asserts unknown roles pass validation. This brings the Anthropic path into parity. Streaming and non-streaming share the same request model, so both modes are covered by one change.
- **Field-tested**: we've been running this exact patch locally in front of Claude Code (Sonnet 5 / Opus 4.8) since it started sending inline `system` messages; the 422s disappeared and conversations flow normally.

## Behavior change (intentional)

`tests/unit/test_routes_anthropic.py::TestMessagesValidation::test_validates_invalid_role` previously asserted unknown role → 422. That assertion enforced the bug, so it is replaced by `test_accepts_non_standard_role` (mirroring the existing OpenAI-side test of the same name and expectation).

## Test coverage

- `tests/unit/test_models_anthropic.py` — new `TestAnthropicMessageRoleAcceptance` (5 tests): user/assistant regression guards, `system` role (primary repro), arbitrary unknown role, and a full `AnthropicMessagesRequest` with an inline system message mid-conversation (exact Claude Code shape).
- `tests/unit/test_converters_anthropic.py` — new end-to-end test `test_inline_system_role_message_is_normalized`: converts a request with an inline system message via `anthropic_to_kiro()` and asserts the Kiro history contains only `userInputMessage`/`assistantResponseMessage` entries and the system content survives as a user message.
- `tests/unit/test_routes_anthropic.py` — route-level tests: `test_accepts_non_standard_role` and `test_accepts_inline_system_role_message` (no 422 at `/v1/messages`).

`pytest -q`: **1698 passed** (full unit + integration suite).

## Related PRs

I'm aware several open PRs address the same 422 (#191, #195, #197, #199, #205, #206, #207, #210, #234...). This one aims to be the minimal, test-covered variant that reuses the gateway's existing normalization rather than introducing new hoisting/merging logic — and it updates the route test that previously locked in the strict behavior. The merge-into-system approach discussed in #190 (comment by @W0n9) is a reasonable semantic refinement that could layer on top of this change later; happy to close this in favor of whichever direction the maintainer prefers.

### DIFF
--- a/kiro/models_anthropic.py
+++ b/kiro/models_anthropic.py
@@ -183,11 +183,15 @@ class AnthropicMessage(BaseModel):
     Message in Anthropic format.
 
     Attributes:
-        role: Message role (user or assistant)
+        role: Message role. Normally "user" or "assistant", but some clients
+            (e.g. Claude Code) occasionally send other values such as "system"
+            for injected context. These are accepted here and coerced to "user"
+            downstream by normalize_message_roles() in converters_core.py,
+            instead of rejecting the whole request with a 422.
         content: Message content (string or list of content blocks)
     """
 
-    role: Literal["user", "assistant"]
+    role: str
     content: Union[str, List[ContentBlock]]
 
     model_config = {"extra": "allow"}
--- a/tests/unit/test_converters_anthropic.py
+++ b/tests/unit/test_converters_anthropic.py
@@ -1499,6 +1499,54 @@ class TestAnthropicToKiro:
         print(f"Current content: {current_content}")
         assert "You are a helpful assistant." in current_content
 
+    def test_inline_system_role_message_is_normalized(self):
+        """
+        What it does: Verifies a request with an inline system-role message
+        converts end-to-end without errors and the message is normalized.
+        Purpose: End-to-end coverage for Issue #190 / #226 - newer Claude Code
+        clients inline role="system" hook context in the messages array.
+        The pipeline (normalize_message_roles) must coerce it to "user" so the
+        Kiro API only ever sees user/assistant history entries.
+        """
+        print("Setup: Request with inline system-role message...")
+        request = AnthropicMessagesRequest(
+            model="claude-sonnet-4-5",
+            max_tokens=1024,
+            messages=[
+                AnthropicMessage(role="user", content="Hello!"),
+                AnthropicMessage(role="assistant", content="Hi!"),
+                AnthropicMessage(role="system", content="Injected hook context"),
+                AnthropicMessage(role="user", content="Continue please"),
+            ],
+        )
+
+        print("Action: Converting to Kiro payload...")
+        with patch(
+            "kiro.converters_anthropic.get_model_id_for_kiro",
+            return_value="claude-sonnet-4.5",
+        ):
+            with patch("kiro.converters_core.FAKE_REASONING_ENABLED", False):
+                result = anthropic_to_kiro(request, "conv-123", "arn:aws:test")
+
+        print(f"Result: {result}")
+        history = result["conversationState"].get("history", [])
+
+        print("Checking history only contains user/assistant entries...")
+        for entry in history:
+            assert (
+                "userInputMessage" in entry or "assistantResponseMessage" in entry
+            ), f"Unexpected history entry keys: {list(entry.keys())}"
+
+        print("Checking system-role content survived as a user message...")
+        history_user_contents = [
+            entry["userInputMessage"]["content"]
+            for entry in history
+            if "userInputMessage" in entry
+        ]
+        assert any(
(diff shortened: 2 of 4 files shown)


Navapon:

ankitcharolia:
@Navapon could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

It supports all necessary endpoints for OpenAI and Anthropic.

PR #260, opened 2026-07-24
feat(models): add Claude Sonnet 5 support

parashuchandra:
## Summary

This PR bundles two related fixes needed to get Claude Sonnet 5 working
end-to-end with Claude Code CLI 2.x through the gateway:

1. **Register `claude-sonnet-5`** as a known model
2. **Fix a request-validation bug** that blocks Claude Code CLI 2.x from
   working with *any* model (not just Sonnet 5) — found while testing #1

### 1. Claude Sonnet 5 support

- `kiro/config.py`: add `claude-sonnet-5` to `FALLBACK_MODELS`, so it's
  recognized as a known model during DNS-failure recovery instead of an
  unverified passthrough.
- `kiro/model_resolver.py`: fix a gap in `normalize_model_name`'s
  no-minor-version pattern. Models without a minor segment (`claude-sonnet-4`,
  `claude-sonnet-5`) only stripped an 8-digit date suffix, not a `latest` or
  trailing-numeric suffix — unlike the with-minor-version pattern, which
  already handled all three. This meant `claude-sonnet-5-latest` failed to
  normalize and incorrectly resolved as unverified passthrough.
- `README.md`: list Claude Sonnet 5 in Available Models.
- `tests/unit/test_config.py`: new `TestClaudeSonnet5Support` — fallback
  registration, name normalization (plain/date/latest suffixes), cache
  resolution (`source=cache`, `is_verified=True`), and family-isolated
  suggestions.

### 2. Accept `role=system` inline in AnthropicMessage

Claude Code CLI 2.x sends messages with `role="system"` inline in the
`messages` array (e.g. injected `<system-reminder>` context), rather than
exclusively via the top-level `system` field. `AnthropicMessage.role` only
accepted `user`/`assistant`, so these requests failed with a 422
`literal_error` — before ever reaching `normalize_message_roles()` in
`converters_core.py`, which already exists to convert unknown roles to
`user` downstream (same lineage as the alternating-roles fix in #64).

Reproduced directly: running `claude -p "..."` against the gateway failed
with:

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Made with ❤️ by [@Jwadow](https://github.com/jwadow)
 
 > 🔒 **Claude Opus 4.5** was removed from the free tier on January 17, 2026. It may be available on paid tiers — check your IDE/CLI model list.
 
+🧠 **Claude Sonnet 5** — Experimental preview with 1M context window. Latest generation, strongest reasoning and coding.
+
 🚀 **Claude Sonnet 4.5** — Balanced performance. Great for coding, writing, and general-purpose tasks.
 
 ⚡ **Claude Haiku 4.5** — Lightning fast. Perfect for quick responses, simple tasks, and chat.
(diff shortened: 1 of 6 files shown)


ankitcharolia:
@parashuchandra could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway and all models are automatically added , no model needed to be updated 

Issue #255, opened 2026-07-19
BUG: Claude Code 2.1.215 embedded system messages fail validation


SimianW:
### Kiro Gateway Version

Current `main` at `a5292ca` (v2.4 development branch)

### What happened?

Claude Code v2.1.215 can place runtime system reminders in `messages[]` using `role: "system"`. Kiro Gateway rejects the request during Pydantic validation because `AnthropicMessage.role` only accepts `user` and `assistant`.

Minimal sanitized request shape:

```json
{
  "model": "<model-id>",
  "messages": [
    {"role": "user", "content": "<redacted>"},
    {
      "role": "system",
      "content": [
        {
          "type": "text",
          "text": "<runtime instructions redacted>",
          "cache_control": {"type": "ephemeral"}
        }
      ]
    }
  ],
  "system": [{"type": "text", "text": "<base system prompt redacted>"}]
}
```

The request never reaches the converter or Kiro runtime. A standard request through the same endpoint succeeds.

Expected behavior: before validating conversation roles, the Anthropic compatibility layer should preserve embedded system content by merging it into the top-level `system` field and remove those entries from `messages[]`. The same normalization should apply to `/v1/messages/count_tokens`.

### Debug Logs

Sanitized validation output:

```text
Validation error (422): [{
  "type": "literal_error",
  "loc": ["body", "messages", 1, "role"],
  "msg": "Input should be 'user' or 'assistant'",
  "input": "system"
}]
```

No prompts, tool definitions, credentials, account identifiers, local paths, or raw debug files are included.

PR #262, opened 2026-07-27
fix: accept inline system messages and merge them into system prompt

### LINKED ISSUE #190: BUG: 422 error when newer Claude Code client inlines `role: "system"` in messages array

aceaura:
## What

Accept `role: "system"` / `role: "developer"` messages inside the `messages` array and merge their text into the system prompt instead of rejecting the request.

Fixes #190, #219, #255.

## Why

Claude Code and Claude Desktop place runtime context (`<system-reminder>` blobs, memory files, session hooks) in a message with `role: "system"` inside `messages`. `AnthropicMessage.role` is `Literal["user", "assistant"]`, so Pydantic rejects the request during FastAPI validation — before any downstream normalization runs — and the client cannot complete a single turn:

```json
422 {"loc": ["body", "messages", 1, "role"],
     "input": "system"}
```

## How

Two parts, because widening the `Literal` alone is not enough:

1. **`models_anthropic.py`** — `AnthropicMessage.role` accepts `system` and `developer`, so the request passes validation. Unknown roles are still rejected, keeping the field constrained.

2. **`converters_anthropic.py`** — new `split_inline_system_messages()` peels those messages off and appends their text to the system prompt; `anthropic_to_kiro()` calls it before `convert_anthropic_messages()`.

Widening the `Literal` on its own would let `normalize_message_roles()` fold the message into a **user turn**, which puts client context into the conversation as if the user had typed it and triggers a synthetic assistant turn to keep roles alternating. Merging into the system prompt instead mirrors what `convert_openai_messages()` already does for OpenAI `role: "system"`, so both APIs now treat inline system messages the same way. Top-level `system` stays first; inline parts are appended in the order the client sent them.

This matches the approach @W0n9 proposed in #190.

## Tests

`tests/unit/test_models_anthropic.py` — `TestAnthropicMessageRoles` (5 tests): each accepted role validates (parametrized), unknown roles still raise, and a full request carrying a `system` message validates.

`tests/unit/test_converters_anthropic.py` — `TestSplitInlineSystemMessages` (7 tests): extraction of `system` and `developer` roles, multiple system messages joined in order, text extracted from content blocks (the `cache_control` shape Claude Code sends), empty content skipped, normal conversations unchanged, empty list. `TestAnthropicToKiroInlineSystem` (4 tests): inline context reaches the Kiro payload, top-level `system` precedes inline text, the system message does not become a history turn, and a request of only system messages raises `ValueError` (surfaced as a client error by `routes_anthropic`).

Full suite: 1708 passed.

### DIFF
--- a/kiro/converters_anthropic.py
+++ b/kiro/converters_anthropic.py
@@ -255,6 +255,45 @@ def extract_tool_uses_from_anthropic_content(content: Any) -> List[Dict[str, Any
     return tool_calls
 
 
+def split_inline_system_messages(
+    messages: List[AnthropicMessage],
+) -> tuple[str, List[AnthropicMessage]]:
+    """
+    Separates inline system messages from the conversation.
+
+    The Anthropic API carries the system prompt in a dedicated top-level
+    field, but several clients (Claude Code, Claude Desktop) also place
+    runtime context in a message with role "system" or "developer" inside
+    the messages array. Kiro only accepts user/assistant turns, so those
+    messages are peeled off here and merged into the system prompt, the same
+    way convert_openai_messages() handles OpenAI system messages.
+
+    Args:
+        messages: List of Anthropic messages, possibly containing inline
+                  system/developer messages
+
+    Returns:
+        Tuple of (inline_system_prompt, conversation_messages)
+    """
+    system_parts = []
+    conversation = []
+
+    for msg in messages:
+        if msg.role in ("system", "developer"):
+            text = convert_anthropic_content_to_text(msg.content)
+            if text:
+                system_parts.append(text)
+        else:
+            conversation.append(msg)
+
+    if system_parts:
+        logger.debug(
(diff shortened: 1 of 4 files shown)


aceaura:



cla-bot[bot] ***@***.***>于2026年7月27日 周一20:03写道：

PR #263, opened 2026-07-27
feat: add /v1/usage endpoint for subscription quota

aceaura:
## What

New `GET /v1/usage` endpoint returning the account's subscription quota: plan name, credits used / remaining, overage cap and rate, and reset date.

```json
{
  "isValid": true,
  "planName": "KIRO PRO MAX",
  "subscriptionType": "Q_DEVELOPER_STANDALONE_PRO_MAX",
  "used": 4595.1,
  "total": 5000.0,
  "remaining": 404.9,
  "unit": "Credits",
  "overageCap": 10000.0,
  "overageRate": 0.04,
  "daysUntilReset": 4,
  "nextDateReset": 1785542400.0,
  "raw": { }
}
```

> Heads-up on process: `CONTRIBUTING.md` asks for feature ideas to be discussed in an issue first, and I opened this as a PR because the implementation was already written and tested. Happy to move the discussion to an issue, rescope it, or drop it if this does not fit the project's direction.

## Why

The gateway already holds everything needed to answer "how much quota is left": credentials, token refresh, and the account's region. Clients that want to display quota — provider dashboards such as cc-switch, status lines, monitoring scripts — currently have to reimplement Kiro token handling and refresh logic just to read a number, or shell out to a sidecar. Exposing it here means one implementation instead of one per client.

The endpoint is read-only, sits behind the same `PROXY_API_KEY` as the rest of the gateway, and adds no work to the request path.

## How

`kiro/routes_usage.py` calls Kiro's `getUsageLimits` with the access token from the first initialized account's auth manager, then flattens the primary quota bucket into a summary. The untouched payload is returned under `raw` so callers can read fields the summary does not surface.

Two details worth noting, both found while testing against a live account:

- **Host** — `getUsageLimits` is only served by the legacy Q hosts. `runtime.{region}.kiro.dev` answers it with `UnknownOperationException`, so the request targets `q.{region}.amazonaws.com` and falls back to the `codewhisperer.{region}.amazonaws.com` alias on a non-200 or transport error.
- **`profileArn`** — never sent. It is optional for most account types, and Enterprise IdC accounts reject the call with 400 when it is present.

The account's own region is used rather than a hardcoded `us-east-1`.

Errors are explicit rather than silently reported as zero usage: 503 when no account is initialized, 502 when every host fails.

## Tests

`tests/unit/test_routes_usage.py`, 14 tests:

- `build_usage_summary` — primary bucket extracted, `raw` preserved, missing breakdown list (free tier returns an empty list), missing plan title defaults, remaining clamped at zero when usage exceeds the limit, `*WithPrecision` fields preferred over the rounded integers.
- `fetch_usage_limits` — bearer token and query params sent, `profileArn` absent, account region used in the URL, fallback host on error status, fallback host on transport error, 502 when all hosts fail.
- `GET /v1/usage` — 401 unauthenticated, 503 with no initialized account, 200 summary for a ready account.

Docs: endpoint added to the README table, test file added to the `tests/README.md` index.

Full suite: 1706 passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,6 +517,7 @@ Leave `VPN_PROXY_URL` empty (default) if you don't need proxy support.
 | `/v1/models` | GET | List available models |
 | `/v1/chat/completions` | POST | OpenAI Chat Completions API |
 | `/v1/messages` | POST | Anthropic Messages API |
+| `/v1/usage` | GET | Subscription quota (credits used / remaining, reset date) |
 
 ---
 
--- a/kiro/routes_usage.py
+++ b/kiro/routes_usage.py
@@ -1,0 +1,149 @@
+# -*- coding: utf-8 -*-
+
+# Kiro Gateway
+# https://github.com/jwadow/kiro-gateway
+# Copyright (C) 2025 Jwadow
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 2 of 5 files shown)


ankitcharolia:
@aceaura could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #267, opened 2026-08-01
fix(docker): set runtime user via compose to match host uid for volume access

facorazza:
Run as the host user so mounted volumes (e.g. kiro-cli SQLite DB) are accessible. Since the kiro user in the image owns /app, this PR ensures the UID matches the host user who owns the files on mounted volumes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,7 @@ services:
       context: .
       dockerfile: Dockerfile
     container_name: kiro-gateway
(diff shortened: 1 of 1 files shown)


facorazza:
I have read the CLA and I don't accept its terms

ankitcharolia:
@facorazza could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

Issue #129, opened 2026-04-12
BUG: RemoteProtocolError mid-stream - Kiro API drops connection during chunked transfer


severity1:
### Kiro Gateway Version

v2.3 (latest main)

### What happened?

The Kiro API intermittently closes the TCP connection mid-stream during chunked transfer encoding, causing `RemoteProtocolError: peer closed connection without sending complete message body (incomplete chunked read)`.

This affects both streaming and non-streaming paths. The gateway handles it correctly (logs the error, returns 500 to client), but the error is not retryable since partial content may have already been sent to the client.

### Error Logs

```
ERROR | kiro.streaming_core:parse_kiro_stream:226 - Error during stream parsing: [RemoteProtocolError] peer closed connection without sending complete message body (incomplete chunked read)
ERROR | kiro.streaming_anthropic:stream_kiro_to_anthropic:525 - Error during Anthropic streaming: [RemoteProtocolError] peer closed connection without sending complete message body (incomplete chunked read)
ERROR | kiro.routes_anthropic:stream_wrapper:390 - HTTP 500 - POST /v1/messages (streaming) - [RemoteProtocolError] peer closed connection without sending complete message body (incomplete chunked read)
```

### Frequency

Observed 2 occurrences in ~4 minutes during normal usage, with 37 successful requests in the same window. Appears intermittent.

### Analysis

- The error originates from `httpx.RemoteProtocolError` during `response.aiter_bytes()` in `parse_kiro_stream()`
- The existing `request_with_retry` logic in `http_client.py` only covers the initial request, not mid-stream failures
- Mid-stream retry is not practical since partial SSE events may have already been sent to the client
- The first-token retry (`stream_with_first_token_retry`) does not help here since the error occurs after tokens have started flowing

### Possible Improvements

1. Downgrade log level from ERROR to WARNING since this is upstream behavior, not a gateway bug
2. Add a specific error message in the SSE error event so clients know it was a connection drop (vs. a gateway bug)
3. Track frequency of these errors and expose via health endpoint or metrics

### Environment

- OS: WSL2 (Linux 5.15.153.1-microsoft-standard-WSL2)
- Client: Claude Code CLI
- Python: 3.12
- httpx: latest

severity1:
**Update**: The frequency of these errors appears to correlate with credit usage. After running clean for over an hour, the errors started occurring frequently (~4 times in 16 minutes) right around 80% credit usage.

[...]

PR #268, opened 2026-08-01
fix(streaming): report OpenAI stream errors in-band instead of raising

zhangw:
## What

`stream_wrapper()` in `chat_completions()` re-raised the exception after `StreamingResponse` had already sent its headers, so Starlette cannot turn it into an error response and aborts the connection instead:

```
ERROR | kiro.streaming_core:stream_with_first_token_retry - [FirstTokenTimeout] All 3 attempts exhausted - model never responded within 15.0s per attempt
ERROR | kiro.routes_openai:stream_wrapper - HTTP 500 - POST /v1/chat/completions (streaming) - [HTTPException] 504: Model did not respond within 15.0s after 3 attempts.
ERROR | Exception in ASGI application
...
  File "starlette/middleware/base.py", line 198, in __call__
    raise app_exc
  File "starlette/_exception_handler.py", line 56, in wrapped_app
    raise RuntimeError("Caught handled exception, but response already started.") from exc

RuntimeError: Caught handled exception, but response already started.
```

Two consequences:

1. The client gets `data: [DONE]` with no content and no reason - it looks like the gateway silently returned an empty answer. The 504 detail never reaches it.
2. Every occurrence prints a full ASGI traceback for what is an upstream condition, not a gateway bug.

`/v1/messages` already handles this correctly (`event: error`, then a graceful end of stream), so only the OpenAI path was affected.

## Why it happens in practice

Any streaming failure after the first byte of the response reaches this path. The easiest way to hit it is the first-token retry loop, because not every model streams reasoning deltas: the non-Claude models exposed by Kiro send nothing at all while they think, so the first byte only arrives together with the answer.

Measured with one heavy request (~270 KB payload, 20 tools, non-Claude model), first-token timeout temporarily raised to 180 s to observe the real latency:

| attempt | upstream headers | first body byte |
|---|---|---|
| 1 | 8.5 s | 38.8 s |
| 2 | 12.4 s | 32.0 s |
| 3 | 9.6 s | 35.2 s |

A Claude model on the same gateway starts emitting `reasoning_content` after ~2-4 s, which is why the default `FIRST_TOKEN_TIMEOUT=15` has been fine so far. With a model that stays silent while reasoning, all three attempts time out and the resulting 504 lands on the broken error path above.

Whether the default timeout should change is a separate discussion - this PR only fixes the error reporting.

Related: #129 (mid-stream `RemoteProtocolError`) hits the same code path on `/v1/chat/completions`, and improvement 2 requested there - telling the client it was a connection drop - now works for the OpenAI API too.

## Change

- New helper `build_sse_error_chunk(exc)` renders an exception as an OpenAI-format error chunk. `HTTPException.status_code` / `.detail` are preserved, anything else is reported as 500.
- Both `stream_wrapper()` variants (account-system branch and legacy branch) now yield the error chunk followed by `[DONE]` and no longer re-raise.

The client now receives the reason:

```
data: {"error": {"message": "Model did not respond within 15.0s after 3 attempts. Please try again.", "type": "kiro_api_error", "code": 504}}

data: [DONE]
```

Server-side logging and debug-log flushing are unchanged - `streaming_error` is still set, so the existing `HTTP 500 - POST /v1/chat/completions (streaming)` access log line and `flush_on_error()` still fire.

## Tests

`tests/unit/test_routes_openai.py`

- `TestBuildSseErrorChunk` - HTTPException status/detail preserved; generic exception reported as 500; exception with an empty message still produces valid JSON.
- `TestStreamingErrorAfterResponseStart` - first-token timeout produces `code: 504` + `[DONE]`; mid-stream failure keeps the chunks already sent and appends `code: 500`; happy path unaffected (no error chunk, exactly one `[DONE]`).

`tests/unit/test_routes_anthropic.py`

- `TestStreamingErrorAfterResponseStart` - parity test asserting `/v1/messages` reports the failure as `event: error` without raising. It passes on `main` as well; it is there to keep the two APIs from drifting apart again.

Verification: 5 of the 7 new tests fail on `main` (the 3 helper tests plus the 2 OpenAI streaming-failure tests, which abort with the `RuntimeError` above) and all 7 pass with this change. Full unit suite: same pass/fail set before and after this change, no new failures.

### DIFF
--- a/kiro/routes_openai.py
+++ b/kiro/routes_openai.py
@@ -86,6 +86,32 @@ async def verify_api_key(auth_header: str = Security(api_key_header)) -> bool:
     return True
 
 
+def build_sse_error_chunk(exc: Exception) -> str:
+    """
+    Build an SSE data chunk describing a streaming failure.
+    
+    Once StreamingResponse has sent its headers, the HTTP status code can no
+    longer be changed: re-raising makes Starlette abort with "Caught handled
+    exception, but response already started" and the client receives a stream
+    with no error information at all. Reporting the error in-band keeps the
+    reason visible to the client (same approach as /v1/messages).
+    
+    Args:
+        exc: Exception raised while streaming. HTTPException status_code and
+             detail are preserved, any other exception is reported as 500.
+    
+    Returns:
+        SSE chunk in OpenAI error format, ready to be yielded
+    """
+    return "data: " + json.dumps({
+        "error": {
+            "message": str(getattr(exc, "detail", None) or exc),
+            "type": "kiro_api_error",
+            "code": getattr(exc, "status_code", 500),
+        }
+    }) + "\n\n"
+
+
 # --- Router ---
 router = APIRouter()
 
@@ -397,11 +423,13 @@ async def chat_completions(request: Request, request_data: ChatCompletionRequest
                                 logger.debug("Client disconnected during streaming (GeneratorExit in routes)")
                             except Exception as e:
                                 streaming_error = e
+                                # Response headers are already sent: report the error
+                                # in-band instead of raising (see build_sse_error_chunk)
                                 try:
+                                    yield build_sse_error_chunk(e)
                                     yield "data: [DONE]\n\n"
                                 except Exception:
-                                    pass
-                                raise
+                                    pass  # Client already disconnected
                             finally:
                                 await http_client.close()
                                 if streaming_error:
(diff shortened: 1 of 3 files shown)


zhangw:

ankitcharolia:
@zhangw could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #269, opened 2026-08-01
fix: harden protocol handling and rate-limit recovery

angelarose210:
## Summary


angelarose210:

ankitcharolia:
@angelarose210 could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #278, opened 2026-09-04
Add claude-auto alias for the auto model

kylerjensen:
## Summary
- Adds `claude-auto` as an additional default entry in `MODEL_ALIASES`, mapping to the underlying `auto` model.

## Why
Claude Desktop rejects model names that lack an Anthropic-y term (e.g. `claude`, `anthropic`, `opus`, `sonnet`, `haiku`). Neither the raw `auto` model ID nor the existing `auto-kiro` alias satisfy that check, so Claude Desktop users can't select the auto model at all. Adding `claude-auto` alongside `auto-kiro` gives Claude Desktop users a usable name while leaving existing aliases and behavior untouched.

## Test plan
- [x] `pytest tests/unit/test_model_resolver.py` — 118 passed
- [x] Full suite `pytest` — 1691 passed

### DIFF
--- a/kiro/config.py
+++ b/kiro/config.py
@@ -245,9 +245,11 @@ HIDDEN_MODELS: Dict[str, str] = {
 #       "gpt-5": "claude-sonnet-4.5"
 #   }
 #
-# Default: {"auto-kiro": "auto"} to avoid Cursor IDE conflict
+# Default: {"auto-kiro": "auto", "claude-auto": "auto"} to avoid Cursor IDE conflict
+# and Claude Desktop's rejection of model names without an Anthropic-y term.
 MODEL_ALIASES: Dict[str, str] = {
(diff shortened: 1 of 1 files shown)


kylerjensen:

ankitcharolia:
@kylerjensen could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

Claude-auto alias **is already added** there 

kylerjensen:
@ankitcharolia thanks for the suggestion! I tried ankitcharolia/kiro-gateway, but noticed that it doesn't handle the Claude Desktop system prompt well since the Kiro CLI already includes a system prompt. Smarter models like Sonnet 5 and Opus 5 disregard the Claude Desktop system prompt that gets included with the `System:` prefix. It also chews through tokens faster due to the duplicated system prompt unfortunately.

PR #279, opened 2026-09-08
feat(models): add claude-opus-5 to fallback model list

sanjayram-a:
## Summary

Adds `claude-opus-5` to `FALLBACK_MODELS` so it's discoverable via
`/v1/models`, and documents it in the README.

## Why

`runtime.kiro.dev` has no `ListAvailableModels` API (AWS limitation),
so the gateway relies on a static `FALLBACK_MODELS` list in
`kiro/config.py` to advertise available models. Claude Opus 5 is now
live on Kiro's backend but wasn't in that list yet.

I confirmed `claude-opus-5` already works via the gateway's
pass-through resolution (unknown models are forwarded to Kiro as-is)
by sending a direct `/v1/messages` request — it responded successfully
even before this change. This PR just makes it show up in `/v1/models`
for discoverability, consistent with how other recent models
(`claude-opus-4.7`, `claude-sonnet-4.6`, etc.) were added.

## Changes

- `kiro/config.py`: add `{"modelId": "claude-opus-5"}` to `FALLBACK_MODELS`
- `README.md`: note Opus 5 availability (mirrors the existing Opus 4.5 callout)
- `tests/unit/test_config.py`: add a regression test asserting
  `claude-opus-5` is present in `FALLBACK_MODELS`

## Testing

Full test suite passes (`pytest`, 1701+ tests).


sanjayram-a:

ankitcharolia:
@sanjayram-a could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

**This Gateway automatically detects all the models using v1/model endpoint so no need to add a model everytime a new model is released**

Issue #159, opened 2026-05-12
feat: expose Kiro credits usage and account pool status via admin endpoint


binsee:
### Problem

When running kiro-gateway with multiple accounts, there is currently no way to proactively monitor Kiro Credits consumption. The only signal available is a `402`/`429` error after an account is exhausted — by which point requests are already failing.

Operators cannot answer basic questions like:
- How many credits have been used this billing period?
- Which accounts are healthy vs. in circuit-breaker cooldown?
- When does the next billing reset happen?

### Proposed Solution

Add a `GET /admin/accounts/usage` endpoint that:

- Calls Kiro's `GET /getUsageLimits` and `POST /GetProfile` APIs per account
- Returns credits used, credits limit, percent used, next reset date, and subscription info
- Exposes gateway-level state: circuit-breaker status (`healthy`/`tripped`) and request counters
- Caches responses (5 min for usage, 24 h for profile) to avoid hammering the Kiro API
- Supports `?account_id=<id>` to filter a single account and `?force_refresh=true` to bypass cache (with a 30 s minimum interval guard per account)
- Returns stale data with `"stale": true` on upstream errors rather than failing hard

The endpoint lives under `/admin/` (not `/v1/`) so it is not accidentally consumed by AI agent clients.

### Example Response

```json
{
  "object": "list",
  "fetched_at": "2026-05-13T10:00:00+00:00",
  "accounts": [
    {
      "account_id": "acct_ed0fe2baf82b",
      "enabled": true,
      "circuit_state": "healthy",
      "stats": {
        "total_requests": 42,
        "successful_requests": 41,
        "failed_requests": 1
      },
      "profile": {
        "profile_name": "my-kiro",
        "status": "ACTIVE"
      },
      "usage": {
        "subscription_title": "KIRO POWER",
        "subscription_type": "Q_DEVELOPER_STANDALONE_POWER",
        "current_usage": 8990.35,
        "usage_limit": 10000.0,
        "percent_used": 89.9,
        "unit": "Credits",
        "currency": "USD",
        "next_reset_at": "2026-06-01T00:00:00+00:00",
        "days_until_reset": 19
      },
      "stale": false,
      "error": null
    }
  ]
}
```

### Implementation

A working implementation is available — PR forthcoming.

PR #135, opened 2026-04-16
fix(anthropic): accurate token estimation for Anthropic API path

leikaiwei:
## Summary

The Anthropic API path (`/v1/messages`) significantly underreports `input_tokens` due to multiple token counting gaps. This PR fixes all of them in a single focused change.

### Problems Fixed

1. **`count_message_tokens` ignores Anthropic content blocks** — Only `text` and `image_url` were handled. `tool_use`, `tool_result`, `image`, and unknown block types were silently skipped, causing token undercount in conversations with tool calls.

2. **`count_tools_tokens` ignores Anthropic flat tool format** — Only OpenAI `{"type": "function", "function": {...}}` format was recognized. Anthropic flat tools (top-level `name`/`description`/`input_schema`) were counted as 4 service tokens only, missing all actual content.

3. **`estimate_request_tokens` called with messages only** — `stream_kiro_to_anthropic` and `collect_anthropic_response` used `count_message_tokens(messages)` instead of `estimate_request_tokens(messages, tools, system)`, completely ignoring tools and system prompt in the fallback estimate.

4. **No support for Anthropic system block list** — `count_system_tokens` did not exist. When system prompt was passed as a list of content blocks (for prompt caching), it was not counted at all.

5. **`context_usage=0%` zeroes out fallback estimate** — `calculate_tokens_from_context_usage` returns `(0, ..., "unknown", ...)` when percentage is 0, but the code unconditionally overwrote the local estimate, setting `input_tokens` to 0.

6. **`AnthropicUsage` drops cache usage fields** — The model lacked `cache_read_input_tokens` and `cache_creation_input_tokens` fields and did not allow extra fields, so upstream cache usage was silently dropped.

7. **Cache usage never forwarded to client** — Neither streaming nor non-streaming responses included `cache_read_input_tokens` / `cache_creation_input_tokens` from upstream Kiro responses.

### Changes

| File | Change |
|------|--------|
| `kiro/tokenizer.py` | Support `tool_use`/`tool_result`/`image` blocks in `count_message_tokens`; support Anthropic flat tool format in `count_tools_tokens`; add `count_system_tokens` for block list format; propagate `apply_claude_correction` in `estimate_request_tokens` |
| `kiro/streaming_anthropic.py` | Accept `request_tools`/`request_system` params; use `estimate_request_tokens` instead of `count_message_tokens`; guard context_usage override with source check; add `_extract_cache_usage_fields` to forward cache usage (supports both camelCase and snake_case keys) |
[...]

PR #280, opened 2026-09-09
fix(parsers): match the upstream metering event so credits are not lost

leikaiwei:
## Problem

`AwsEventStreamParser.EVENT_PATTERNS` looks for `{"usage":`, but the upstream billing event is:

```json
{"unit":"credit","unitPlural":"credits","usage":0.02730794364842454}
```

`usage` is the third key, so `buffer.find('{"usage":')` never matches it. Consequences:

- `metering_data` in `streaming_openai.py` stays `None`, so the existing `final_chunk["usage"]["credits_used"] = metering_data` has never produced a field. I hit the OpenAI endpoint three times in a row and `credits_used` did not appear once.
- `_extract_cache_usage_fields(event.usage)` in `streaming_anthropic.py` never runs either, so the cache usage forwarding added in #135 has never taken effect.
- `stream_completed_normally = received_usage or received_context_usage` effectively has one leg: `received_usage` is always `False`.

I dumped the raw upstream chunks with `DEBUG_MODE=all` over several requests, including tool-call rounds. The only event keys the current API version sends are `{"content"`, `{"contextUsagePercentage"`, `{"unit"`, `{"name"` and `{"input"`. `{"usage":{...}}` does not appear at all.

The field name in the AWS SDK really is `usage` (`MeteringEvent { usage: f64, unit: String }`), which is probably where the pattern came from, but `unit` is serialized first.

## Fix

A separate `metering` event type rather than reusing `usage`:

- `usage` carries cache token counts (a dict), `metering` carries credits (a float). Reusing one type would let them overwrite each other if upstream ever starts sending `{"usage":{...}}`, and the symptom would be a credit value silently turning into a dict.
- The pattern matches `{"unit":` only, and the unit value is checked in `_process_metering_event`. Matching `{"unit":"credit"` would break silently on one extra space in the upstream JSON. An unknown unit logs a warning instead of being dropped quietly, since it would mean upstream changed how it bills.
- Credits are accumulated, not overwritten. One downstream request can trigger several upstream calls when tools are used: I captured a single `/v1/messages` request with two credit events, 0.0930 + 0.5290, so keeping only the last one undercounts by 15%.
- Both APIs and both streaming and non-streaming paths are covered.
- Each request logs `[Credit] req=<id> model=<model> credits=<float> calls=<n>`.

`usage.credits_used` on the OpenAI endpoint is behind a new `EXPOSE_CREDITS_USED` env var, default off, so that fixing the parser does not by itself start adding a non-standard field to responses. Say the word and I will flip the default to on.

The Anthropic path only accumulates and logs; its payload is unchanged.

One note for whoever touches this later: credits are not dollars, and the field must not be renamed to `cost`. Downstream proxies such as LiteLLM have an explicit `cost` field on their usage model and would treat the value as a real charge.

## Behaviour change

`received_usage` in `streaming_openai.py` starts being `True`, so truncation detection becomes slightly less eager. That restores what the condition was written to do. In every capture I took, `contextUsagePercentage` and the metering event arrive together, so the verdict rarely differs in practice, and receiving a metering event is itself a strong signal that upstream finished and billed the request. The Anthropic path does not consult metering for truncation detection at all, so it is unaffected.

## Tests

`pytest tests/unit`: 1594 passed before, 1602 after, no failures.

New parser tests: a single event, two events in one stream, an event split across chunks, an unknown unit (dropped without swallowing the next event), non-numeric usage, and content that itself contains an escaped `{\"unit\":\"kg\"}`.

New OpenAI streaming tests: the field is absent by default, credits are summed when the flag is on, and a cache-usage dict is never reported as credits.

Separately, I ran the old and the new pattern table side by side over 7 streams (a real capture with binary frame headers, two-round tool calls, escaped `unit` inside content, a chunk split in the middle of the event, an unknown unit, thinking plus signature events, and metering only), asserting that every non-metering event is identical before and after. That is the property I actually cared about: no side effects on existing parsing.

Related: #159 asks for credit visibility through `getUsageLimits`. That endpoint has no per-model breakdown (its `usageBreakdown` and `totalUsage` are null and `ResourceType` has no model dimension), whereas the metering event gives the exact credits for the request that just ran.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -282,6 +282,22 @@ PROXY_API_KEY="my-super-secret-password-123"
 # TRUNCATION_RECOVERY=true
 
 # ===========================================
+# BILLING
+# ===========================================
+
+# Expose the credits billed by Kiro as "usage.credits_used" on the OpenAI endpoint.
+# Credits are ALWAYS written to the log regardless of this setting:
+#   [Credit] req=<id> model=<model> credits=<float> calls=<n>
+# This flag only controls whether they are also added to the response payload,
+# which is a non-standard field for OpenAI clients.
+#
+# Credits are NOT dollars. Never rename this field to "cost" - downstream proxies
+# such as LiteLLM treat "cost" as a real charge and would mis-bill.
+#
+# Default: false
+# EXPOSE_CREDITS_USED=false
+
+# ===========================================
 # LOGGING
 # ===========================================
 
--- a/docs/en/ARCHITECTURE.md
+++ b/docs/en/ARCHITECTURE.md
@@ -329,7 +329,8 @@ Advanced AWS SSE format parser with support for:
 | `tool_start` | Start of tool call (name, toolUseId) |
 | `tool_input` | Continuation of input for tool call |
 | `tool_stop` | End of tool call |
-| `usage` | Credit consumption information |
+| `usage` | Cache token usage (dict; not sent by current upstream versions) |
+| `metering` | Credits billed for the request (`{"unit":"credit",...,"usage":<float>}`) |
 | `context_usage` | Context usage percentage |
 
 #### Helper Functions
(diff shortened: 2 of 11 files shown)


ankitcharolia:
@leikaiwei could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #281, opened 2026-09-15
Update README.md

ChowRex:
更新增加属主修改内容说明


ankitcharolia:
@ChowRex could you give a try to this gateway: https://github.com/ankitcharolia/kiro-gateway

PR #285, opened 2026-09-20
feat(models): auto-discover working-but-unadvertised models

guanbear:
## Problem

On the runtime endpoint (`runtime.<region>.kiro.dev`) there is no `/ListAvailableModels` API, so `/v1/models` is served entirely from the static `FALLBACK_MODELS` list. Models released after that list was last updated (currently `claude-opus-5`, `claude-sonnet-5`, `claude-opus-4.8`, the `gpt-5.6` family) work fine when requested directly — the account manager even "learns" them into its routing map on success — but they never appear in `/v1/models`, which breaks every client that relies on the endpoint for model discovery.

## Approach

Extend the existing dynamic-learning behavior: when a request with a model ID that is not in the cache **succeeds**, register it in the model cache so it becomes visible.

- `ModelInfoCache.add_discovered_model()` — registers the model (marked `_discovered`, default token limits)
- `AccountManager.report_success()` — registers into the account's cache (account-system mode; shared by OpenAI and Anthropic routes)
- legacy single-account mode — same registration right after a 200 from Kiro in both route files

No probing, no extra upstream requests, no configuration, no new dependencies: a model appears in `/v1/models` the first time any client uses it successfully. Registration is idempotent; entries live for the process lifetime (consistent with the existing in-memory model cache).

## Verification

Unit test added (`test_report_success_discovers_new_model`): account initialized with the standard fixture, `report_success()` with `claude-sonnet-5` → model appears in `get_all_available_models()`, is valid in the cache, and a second call does not duplicate it. Passes locally.

End-to-end (2026-09-20, IAM Identity Center account via kiro-cli SQLite, `runtime.us-east-1.kiro.dev`, legacy mode):

| step | result |
|---|---|
| `GET /v1/models` | 13 models, no `claude-opus-5` |
| `POST /v1/chat/completions` with `claude-opus-5` | 200 OK |
| `GET /v1/models` again | 14 models incl. `claude-opus-5`; log: `Discovered new model via successful request: claude-opus-5` |

## Notes

- This makes the model list self-maintaining instead of requiring a release every time Kiro ships a model. It complements static additions: statically listed models appear immediately at startup, discovered models after first successful use.
- Alternative/sibling PRs #242, #279 and part of #260 add specific IDs statically; this mechanism makes that class of PR unnecessary going forward.
- Note for README (happy to follow up): Claude-family models require reaching the runtime from a non-CN IP (`VPN_PROXY_URL`); a direct CN connection gets 400 "insufficient subscription level".
- Test suite: 4 tests in `test_routes_anthropic.py` (TruncationRecovery group) fail when run in the multi-file combination on **main** as well — pre-existing cross-file pollution, unrelated to this change (single-file runs pass; my branch shows the identical 4 failures + this PR's new test passing).


guanbear:
> Thanks for the PR! 🎉
> 
> 
> 
> Please reply once with:
> 
> ```
> ```
> 

guanbear:

